PR #4, opened 2022-01-19
Solution

Krimat:


denyslipin (approved):
Nice job! Your explanations of ValueErrors in both classes is great.

app/main.py
```diff
@@ -1,8 +1,85 @@
...
+            self,
+            distance_from_city_center,
+            clean_power,
+            average_rating,
+            count_of_ratings):
+        if distance_from_city_center > self.max_distance_from_center or\
+                distance_from_city_center < self.min_distance_from_center:
+            raise ValueError(
+                f'distance_from_city_center is {distance_from_city_center},'
+                f' which is out of bonds{self.min_distance_from_center}'
```
IvanRamyk, line 45: Nice messages!
But you probably have a typo in the word "bo**u**nds"

app/main.py
```diff
@@ -1,8 +1,85 @@
...
+    max_comfort_class = 7
+    min_comfort_class = 1
+    max_clean_mark = 10
+    min_clean_mark = 1
+
+    def __init__(self, comfort_class, clean_mark, brand):
+        if comfort_class > self.max_comfort_class \
+                or comfort_class < self.min_comfort_class:
+            raise ValueError(
+                f'comfort_class value is {comfort_class},'
```
IvanRamyk, line 11: We use double quotes by default in this course. So it would be better to use them in error messages. 
Actually, single quotes are usually fine but It is important to stick to one style

IvanRamyk (changes requested):
Nice job!
waiting for small fixes)

MasterpieceElbow (approved):

PR #5, opened 2022-01-20
'Solution'

liuda-chuzdiuk:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        fit_cars = []
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                fit_cars.append(car)
+
+        for fit_car in fit_cars:
+            self.wash_single_car(fit_car)
+
+        return round(income, 1)
```
Krimat, line 27: Avoid using magic numbers

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+    def serve_cars(self, cars: list):
+        income = 0
+        fit_cars = []
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                fit_cars.append(car)
+
+        for fit_car in fit_cars:
+            self.wash_single_car(fit_car)
```
Krimat, line 25: Second 'for' loop is redundant

Krimat (approved):

app/main.py
IvanRamyk: You actually can wash cars in the previous loop, but it is still OK

IvanRamyk (approved):
Well done!

PR #6, opened 2022-01-20
'Solution'

serhii56465:


alix-mak (approved):
In my opinion your solution is fine

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars):
+        resultlist = []
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                resultlist.append(self.calculate_washing_price(car))
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                self.wash_single_car(car)
```
MasterpieceElbow, line 23: Why do you have 2 similar loops? You can calculate price and wash in one loop

MasterpieceElbow (changes requested):

app/main.py
serhii56465: I do it to avoid changing car.clean_mark that I use for self.calculate_washing_price(car). Thant's why at first I calculate with old value of car.clean_mark and THEN I use self.wash_single_car(car) where car.clean_mark changes

serhii56465: I understood.

app/main.py
```diff
@@ -1,8 +1,39 @@
...
-    pass
+    def __init__(self, distance_from_city_center, clean_power,
+                 average_rating, count_of_ratings):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars):
+        resultlist = []
```
IvanRamyk, line 17: You don't actually need a list to store all prices. You can calculate sum without it

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #13, opened 2022-01-20
Solution

IgorOlifirenko:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def calculate_washing_price(self, car):
+        difference = self.clean_power - car.clean_mark
+        division = self.average_rating / self.distance_from_city_center
+        return round(car.comfort_class * difference * division, 1)
+
+    def wash_single_car(self, car):
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+        return car
```
phantom-terminal, line 44: it's not a necessary return. ↑

phantom-terminal (changes requested):

phantom-terminal (approved):

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                 count_of_ratings: int
+                 ):
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list):
+        amount = []
```
IvanRamyk, line 27: You don't need to store all prices in the list to calculate sum

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        difference = self.clean_power - car.clean_mark
+        division = self.average_rating / self.distance_from_city_center
+        return round(car.comfort_class * difference * division, 1)
+
+    def wash_single_car(self, car):
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, mark):
+        new_average = self.average_rating * self.count_of_ratings + mark
```
IvanRamyk, line 46: It seems, that it's not a `new_average`, but new sum of marks.
Could you change this name?

IvanRamyk (changes requested):

app/main.py
IgorOlifirenko: Done

IvanRamyk (approved):
Well done!

PR #15, opened 2022-01-20
Solution

vldtsw:


Oleksandr563487953 (approved):

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+    def serve_cars(self, cars_list: list):
+        income = sum([self.calculate_washing_price(car) for car in cars_list
+                      if car.clean_mark < self.clean_power])
+        for car in cars_list:
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car):
+        clean_dif = self.clean_power - car.clean_mark
+        clean_ratio = self.average_rating / self.distance_from_city_center
```
MasterpieceElbow, line 26: There is not sense to make individual variable

MasterpieceElbow (changes requested):

app/main.py
Oleksandr563487953: I thought it is more clean to understand, so i approved it.

vldtsw: I made it to avoid "E501 line too long" :(

IvanRamyk (approved):
Good job!

PR #19, opened 2022-01-21
Solution

Geksus:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                 average_rating: float, count_of_ratings: float):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list):
+
+        income = 0
+        fit_cars = []
```
cloudprogramm, line 19: fit_cars is gets cars which fit for condition but it variable never use.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                fit_cars.append(car)
```
cloudprogramm, line 24: You can remove code with fit-cars and it's still working, therby you will save time and memory.

cloudprogramm (approved):
Good job!

app/main.py
MasterpieceElbow: Yes, variable 'fit_cars' is useless.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, other):
```
MasterpieceElbow, line 29: You always takes car as 'other', so it is better to rename 'other' to 'car'

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            ((other.comfort_class * (self.clean_power - other.
+                                     clean_mark) * self.average_rating) / self.
+             distance_from_city_center),
+            1)
+
+    def wash_single_car(self, other):
+        if self.clean_power > other.clean_mark:
+            other.clean_mark = self.clean_power
+
+        return other
```
MasterpieceElbow, line 40: The same, rename 'other' to 'car' and don't return anything.

MasterpieceElbow (changes requested):

app/main.py
Geksus: Fixed

app/main.py
Geksus: Fixed

app/main.py
Geksus: Fixed

app/main.py
Geksus: Fixed

MasterpieceElbow (approved):

PR #21, opened 2022-01-21
Solution

nikon4uk:


vldtsw (changes requested):
You have a lot of flake8 issues. Please run `flake8 app` in your IDE terminal and fix them

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+            product = other.comfort_class * difference * self.average_rating
+            return product / self.distance_from_city_center
+        else:
+            raise TypeError(f"'{type(other)}' not supported type. Requires 'Car'")
+
+    def wash_single_car(self, other: Car):
+        if isinstance(other, Car):
+            other.clean_mark = self.clean_power
+        else:
+            raise TypeError(f"{type(other)} type not supported. Requires 'Car'")
```
Abnormaltype, line 33: If I want to use this method on objects what do I receive?
you need condition   `if self.clean_power > other.clean_mark:`

Abnormaltype (changes requested):

app/main.py
nikon4uk: Yes, I understand. I would generally make this method static. But then it will not fit into the condition of the problem. I assumed that the car will only be washed if they pay for it. In another case, I will repeat the same condition, or I will have to go through the list of machines twice.

Abnormaltype: I understand but the task says that the function must be implemented this way

Abnormaltype (approved):

MasterpieceElbow (approved):

PR #26, opened 2022-05-03
car_washing_station v1

SairusQw:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        return round(price / self.distance, 1)
+
+    def wash_single_car(self, car):
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+        return car
+
+    def rate_service(self, rating):
+        self.count_of_ratings += 1
+        full_rating = self.average_rating * (self.count_of_ratings - 1)
```
IvanRamyk, line 41: If you swap the first two lines of the method It'll become more readable :) In that case, you shouldn't subtract one from the rating count.
```suggestion
        full_rating = self.average_rating * self.count_of_ratings
```

IvanRamyk (changes requested):

Abnormaltype (changes requested):
All tests should are passed

IvanRamyk (approved):
Good job!

PR #32, opened 2022-05-15
'Solution'

Magus6784:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def wash_single_car(self, car):
+        car.clean_mark = self.clean_power
```
TPolina, line 29: add check that clean_mark < clean_power. If you run this method separately it will not work correctly

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def calculate_washing_price(self, car: Car):
+        return round(
+            (car.comfort_class * (self.clean_power - car.clean_mark
+                                  ) * self.average_rating
+             ) / self.distance_from_city_center, 1
+        )
```
TPolina, line 36: bad style, try to make calculation more readable

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def rate_service(self, rate):
+        self.average_rating = round(
+            ((self.average_rating * self.count_of_ratings) + rate
+             ) / (self.count_of_ratings + 1), 1)
```
TPolina, line 41: same here

TPolina (changes requested):

app/main.py
Magus6784: Fixed

TPolina (approved):
Well done!

PR #33, opened 2022-05-15
'Solution'

Serhii-Leonenko:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                         / self.distance_from_city_center, 1)
+
+    def serve_cars(self, list_of_cars: list):
+        income = 0
+        for car in list_of_cars:
+            if self.can_wash_car(car):
+                income += (car.comfort_class
+                           * (self.clean_power - car.clean_mark)
+                           * self.average_rating
+                           / self.distance_from_city_center)
```
TPolina, line 42: don't repeat yourself: use method you wrote

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                car.clean_mark = self.clean_power
```
TPolina, line 43: same here

TPolina (changes requested):

TPolina:
fix flake8

app/main.py
Serhii-Leonenko: But, if i don't mistake, it will be the best practice soon, or no ?)
https://prnt.sc/0nwHZjOTU9Bx

app/main.py
TPolina: and here use `wash_single_car` method

TPolina (changes requested):
please, fix remarks in the comments

app/main.py
Serhii-Leonenko: aaaa, ok

app/main.py
Serhii-Leonenko: yes, i understood you

app/main.py
Serhii-Leonenko: Ok, i will try)

Serhii-Leonenko: Yes, you are right, unfortunately  i didn't think about that i can not to round the result of  serve_cars

TPolina (approved):

PR #34, opened 2022-05-15
Solution

MykolaIF:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round(
+            (self.average_rating * (self.count_of_ratings - 1) +
+             + rate) / self.count_of_ratings, 1
+        )
```
Abnormaltype, line 46: You can calculate `self.average_rating `first and then increment `self.count_of_ratings.`
In that case, you get rid of -1 in the expression, which could be a bit confusing.

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def calculate_washing_price(self, car: Car):
+        price = (
+            car.comfort_class * (self.clean_power - car.clean_mark) * self.
+            average_rating / self.distance_from_city_center
+        )
+
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Abnormaltype, line 39: please implement this 
`wash_single_car` - method, that washes a single car, so it should have `clean_mark` equals wash station's `clean_power`, if `wash_station.clean_power` is greater than `car.clean_mark`

Abnormaltype (changes requested):

PR #38, opened 2022-05-16
make solution to hw4

BohdanKryven:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        return round(price, 1)
+
+    def calculate_washing_price(self, car: Car):
+        return round(
+            car.comfort_class * abs(
+                self.clean_power - car.clean_mark
+            ) * self.average_rating / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car):
+        car.clean_mark = self.clean_power
```
TPolina, line 36: add check that clean_mark < clean_power. When you run this method separately it doesn't work correctly

TPolina (changes requested):

TPolina (approved):

PR #39, opened 2022-05-16
'Solution'

darten360:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        price = cost / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car):
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate):
+        rating = self.average_rating * self.count_of_ratings + rate
+        count = self.count_of_ratings + 1
```
TPolina, line 48: better write here self.count_of_ratings += 1 and use it not to perform double calculation

TPolina (changes requested):

PR #41, opened 2022-05-16
Solution

RomaTsybulia:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+        self.all_rate = average_rating * count_of_ratings
+
+    def serve_cars(self, all_cars):
+        income = 0
+
+        for car in all_cars:
+            if car.clean_mark < self.clean_power:
+                print(car.clean_mark)
```
TPolina, line 28: redundant line

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        return income
+
+    def calculate_washing_price(self, car):
+        cc = car.comfort_class
+        cp = self.clean_power
+        cm = car.clean_mark
+        ar = self.average_rating
+        dfcc = self.distance_from_city_center
+
+        return f'{(cc * (cp - cm) * ar / dfcc):.1f}'
```
TPolina, line 40: no need to return string, return just the result of calculation

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            distance_from_city_center,
+            clean_power, average_rating,
+            count_of_ratings
+    ):
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+        self.all_rate = average_rating * count_of_ratings
```
TPolina, line 21: bad idea to assign temporary calculation to object attributes, as they are suppose to describe properties of the object not just help you in one of the methods

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                income += float(self.calculate_washing_price(car))
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car):
+        cc = car.comfort_class
+        cp = self.clean_power
+        cm = car.clean_mark
+        ar = self.average_rating
+        dfcc = self.distance_from_city_center
```
TPolina, line 38: not the best way to make your code fit in one line, better split calculation into separate actions

TPolina (changes requested):

app/main.py
```diff
@@ -1,8 +1,48 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(
+            self,
+            distance_from_city_center,
+            clean_power, average_rating,
```
TPolina, line 13: choose one way: write all parameters in one line or each from new line

app/main.py
TPolina: still not fixed

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car):
+        car_class = car.comfort_class
+        clean_difference = self.clean_power - car.clean_mark
+        rating = self.average_rating
+        distance = self.distance_from_city_center
+
+        return f'{(car_class * clean_difference * rating / distance):.1f}'
```
TPolina, line 37: no need to return string, return just the result of calculation like you do in `rate_service`

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, all_cars):
+        income = 0
+        for car in all_cars:
+            if car.clean_mark < self.clean_power:
+                print(car.clean_mark)
+                income += float(self.calculate_washing_price(car))
```
TPolina, line 27: here you can use the result of calculation, that you return, without converting it

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car):
+        car_class = car.comfort_class
+        clean_difference = self.clean_power - car.clean_mark
+        rating = self.average_rating
+        distance = self.distance_from_city_center
```
TPolina, line 35: still my opinion it it's better to split calculation into separate actions instead of renaming variables

TPolina (changes requested):

app/main.py
TPolina: remove `print` statement because it is not needed

app/main.py
TPolina: no need to convert into `float` because your value already has `float` type

TPolina (changes requested):

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, all_cars):
+        income = 0
+        for car in all_cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car
+            self.wash_single_car(car)
```
TPolina, line 32: better call this function inside if statement

TPolina (changes requested):
fix syntax errors. Don't forget to run flake8 and pytest before pushing your solution

TPolina (approved):
Nice)

PR #43, opened 2022-05-16
'Solved'

SergeyDev1996:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                profit += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return profit
+
+    def calculate_washing_price(self, car: Car):
+        price = (
+            car.comfort_class * (self.clean_power - car.clean_mark) * self.
+            average_rating / self.distance_from_city_center
+        )
```
TPolina, line 36: bad long code splitting

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
TPolina, line 41: add check that clean_mark < clean_power. When you run this method separately it doesn't work correctly

TPolina (changes requested):
fix flake8

TPolina (approved):

PR #44, opened 2022-05-17
'Solution'

daria-niklaevna:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        multipl = car.comfort_class * difference * self.average_rating
+        result = multipl / self.distance
+        return round(result, 1)
+
+    def serve_cars(self, cars):
+        price = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
```
TPolina, line 35: better use method you wrote

TPolina (changes requested):

app/main.py
TPolina: not fixed. You should use wash_single_car here

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        return price
+
+    def wash_single_car(self, car):
+        if self.clean_power > car.clean_mark:
+            self.wash_single_car(car)
```
TPolina, line 40: it looks like infinite cycle :)

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def rate_service(self, rate):
+        rates = self.average_rating * self.count_of_ratings + rate
+        count = self.count_of_ratings + 1
```
TPolina, line 44: better write here `self.count_of_ratings += 1` and use it not to perform double calculation

TPolina (changes requested):

PR #50, opened 2022-05-17
It seems to work correctly

Flashmobber:


app/main.py
```diff
@@ -1,8 +1,48 @@
 class Car:
-    # write your code here
-    pass
+    comfort_class: int
+    clean_mark: int
+    brand: str
```
TPolina, line 4: what is the purpose of these lines?

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def __init__(self, c_class: int, c_mark: int, car_brand: str):
```
TPolina, line 6: use more descriptive names of parameters. The person who will use your class will not understand what he should pass to it

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.brand = car_brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    distance_from_city_center: float
+    clean_power: int
+    average_rating: float
+    count_of_ratings: int
```
TPolina, line 16: same here

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def __init__(self, distance: float, c_power: int,
+                 rating: float, count_of_votes: int):
```
TPolina, line 19: some of these parameters' names are not good as well

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.distance_from_city_center = distance
+        self.clean_power = c_power
+        self.average_rating = rating
+        self.count_of_ratings = count_of_votes
+
+    def serve_cars(self, washes: list) -> float:
+        def calculation(car):
+            result = self.calculate_washing_price(car)
+            self.wash_single_car(car)
+            return result
```
TPolina, line 29: better call this method `serve_one_car` and write it outside `serve_cars` method and then just use it here

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+        return sum(calculation(car) for car in washes
+                   if self.clean_power > car.clean_mark)
+
+    def calculate_washing_price(self, car: Car):
+        return round(car.comfort_class * (
+                     self.clean_power - car.clean_mark) * (
+                     self.average_rating / self.distance_from_city_center), 1)
```
TPolina, line 37: not really good long code splitting

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def wash_single_car(self, car: Car):
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+        return car
+
+    def rate_service(self, vote: int):
+        vt = round((self.average_rating * self.count_of_ratings + vote
+                    ) / (self.count_of_ratings + 1), 1)
```
TPolina, line 46: same here

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.average_rating = round(vt, 1)
```
TPolina, line 47: you use `round` twice

TPolina (changes requested):

TPolina (changes requested):

TPolina (approved):

PR #56, opened 2022-05-17
Solution

Chardona:


app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars):
+        clients = []
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                clients.append(car)
+        income = self.calculate_washing_price(clients)
+        return income
```
TPolina, line 27: iterate through each car, calculate the price for it's washing with the help of `calculate_washing_price method` and then return sum of prices

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+
+    def rate_service(self, rate: int):
+        nominator = self.average_rating * self.count_of_ratings + rate
+        denominator = self.count_of_ratings + 1
+        self.average_rating = round(nominator / denominator, 1)
+        self.count_of_ratings += 1
+
+    def set_clean_status(self, car):
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
```
TPolina, line 55: there is no need in this method. just implement `wash_single_car` and use it

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+                income += nom / self.distance_from_city_center
+                self.set_clean_status(client)
+            return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        self.set_clean_status(car)
+
+    def rate_service(self, rate: int):
+        nominator = self.average_rating * self.count_of_ratings + rate
+        denominator = self.count_of_ratings + 1
```
TPolina, line 49: better write here self.count_of_ratings += 1 and use it not to perform double calculation

TPolina (changes requested):

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+    def serve_cars(self, cars):
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car):
+        income = 0
```
TPolina, line 27: no need for this line: you perform calculation only once and can just return a result instead of adding it to 0 and only then return

TPolina (changes requested):

TPolina (approved):
Great!

PR #70, opened 2022-05-31
Final

CEHR2005:


app/main.py
```diff
@@ -1,8 +1,36 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center, clean_power,
+                 average_rating, count_of_ratings):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, ls: list):
```
Abnormaltype, line 16: `ls` - it's no informative name for the variable

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        inc = 0
```
Abnormaltype, line 17: same for `inc`

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        for i in ls:
```
Abnormaltype, line 18: same for `i`

Abnormaltype (changes requested):
Suggestion:
if you use type annotations, please write type hinting in all methods

PR #71, opened 2022-06-06
Solution

andrii-padalko:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 clean_power,
+                 average_rating,
+                 count_of_ratings):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car):
+        clean_dif = abs(car.clean_mark - self.clean_power)
```
IvanRamyk, line 20: Why do you use abs here?
It seems that in this case If the car is cleaner than the clean mark of the station a driver should pay for service

IvanRamyk: I'd choose the approach where this method returns zero if the car is already clean

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        rat_dist = self.average_rating / self.distance_from_city_center
```
IvanRamyk, line 21:
```suggestion
        rate_dist = self.average_rating / self.distance_from_city_center
```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        return round(
+            car.comfort_class * clean_dif * rat_dist
+            , 1
```
IvanRamyk, line 24: Maybe place store this value  `car.comfort_class * clean_dif * rat_dist` in the variable to avoid return statement three lines long

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #72, opened 2022-06-06
Solution

artemkazakov947:
Solution with tears

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class, clean_mark, brand):
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center,
+                 clean_power,
+                 average_rating,
+                 count_of_ratings):
+        self.distance_from_sity_center = distance_from_city_center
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+                 clean_power,
+                 average_rating,
+                 count_of_ratings):
+        self.distance_from_sity_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def rate_service(self, rate):
+        a = self.average_rating * self.count_of_ratings + rate
```
IvanRamyk, line 20: a isn't a good variable name. Please, rename it

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.average_rating = round(a / self.count_of_ratings + 1, 1)
+        self.count_of_ratings += 1
+
+    def calculate_washing_price(self, car):
+        b = car.comfort_class * (self.clean_power - car.clean_mark)
+        c = b * self.average_rating / self.distance_from_sity_center
```
IvanRamyk, line 26: please, rename variable c
b is not a good name too, but I understand that it stores an intermediate result so it's difficult to choose an appropriate name for it

IvanRamyk (changes requested):

artemkazakov947:
a was replaced by add_new_rate. b was replaced by var and I added comment inside the function. c was replaced by price.

PR #73, opened 2022-06-07
Kazakov

artemkazakov947:
Variable a was replaced with add_new_rate. Variable b was replaced with var and I added comment inside the function. Variable c was replaced with price.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class, clean_mark, brand):
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def rate_service(self, rate):
+        new_rate = self.average_rating * self.count_of_ratings + rate
+        self.average_rating = round(new_rate / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
+
+    def calculate_washing_price(self, car):
+        # var is car's comfort class * difference between
```
IvanRamyk, line 25: please, remove comments. It's better to write self-descriptive code

IvanRamyk (changes requested):

PR #78, opened 2022-06-16
Solution

Vlad-Peresta:


app/main.py
```diff
@@ -1,8 +1,37 @@
...
+    def serve_cars(self, list_of_cars):
+        income = []
+        for car in list_of_cars:
+            if car.clean_mark < self.clean_power:
+                income.append(self.calculate_washing_price(car))
+            self.wash_single_car(car)
+        return sum(income)
+
+    def rate_service(self, rate):
+        numerator = self.average_rating * self.count_of_ratings + rate
```
Abnormaltype, line 35: `numerator`? Maybe would be better `new_rating`?)

Abnormaltype (changes requested):

app/main.py
Vlad-Peresta: with 'numerator' line consist 70 characters

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        for car in list_of_cars:
+            if car.clean_mark < self.clean_power:
+                income.append(self.calculate_washing_price(car))
+            self.wash_single_car(car)
+        return sum(income)
+
+    def rate_service(self, rate):
+        new_rating = self.average_rating * self.count_of_ratings + rate
+        self.average_rating = round(new_rating / (self.count_of_ratings + 1),
+                                    1)
```
Abnormaltype, line 37: very strange indentation

Abnormaltype: Would be better `self.average_rating = round(new_rating / (self.count_of_ratings + 1), 1)`

Abnormaltype (changes requested):

app/main.py
Abnormaltype: or  `self.count_of_ratings += 1` and   `self.average_rating ` swap places and problems will not be

Abnormaltype (changes requested):

app/main.py
Vlad-Peresta: thank you very much for this idea) it will be the best variant

PR #85, opened 2022-06-23
Solution

anatomst:


app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        return result
+
+    def calculate_washing_price(self, car):
+        price = car.comfort_class * (self.clean_power - car.clean_mark)\
+            * self.average_rating / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car):
+        if car.clean_mark < self.clean_power:
+            return car.comfort_class * (self.clean_power - car.clean_mark)\
```
IvanRamyk, line 31: This method shouldn't return anything, just to change car's clean mark

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars):
+        result = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                result += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
```
IvanRamyk, line 21: Use `wash_single_car` method here to change clean mark

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        return result
+
+    def calculate_washing_price(self, car):
+        price = car.comfort_class * (self.clean_power - car.clean_mark)\
```
IvanRamyk, line 25: if `clean_mark` is greater than `clean_power` the method should return 0, not a negative number

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #89, opened 2022-07-02
MYSolution

shevchenko-anastasiia:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: float,
+                 average_rating: float,
+                 count_of_ratings: int):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car):
```
IvanRamyk, line 17:
```suggestion
    def serve_cars(self, cars):
```
It's better to name lists in plural

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        income = 0
+        for new_car in car :
+            if new_car.clean_mark <= self.clean_power :
```
IvanRamyk, line 20: Why the car is new?)
Also a bit confusing name

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #95, opened 2022-07-13
cost_calculation

Oleksy-Syrota:


Abnormaltype (changes requested):
all test should are pass

PR #99, opened 2022-07-16
My Solution

YanPurdenko:


tests/test_main.py
```diff
@@ -1,103 +0,0 @@
-import pytest
```
IvanRamyk, line 1: Why did you delete all tests?

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_mark = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car: Car):
```
IvanRamyk, line 28:
```suggestion
    def serve_cars(self, cars: Car):
```
I believe, this is a list of the cars

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        income = 0
+        for i in car:
```
IvanRamyk, line 30: Please, choose another name for the loop variable

IvanRamyk (changes requested):

PR #153, opened 2022-09-06
Develop

musakovskiy:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def serve_cars(self, cars: list):
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                self.wash_single_car(car)
+        return self.income
+
+    def calculate_washing_price(self, car):
+        price = car.comfort_class * \
+            (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
```
DmytroSvirsa, line 30: split to different variables to make it more readable

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        car.clean_mark = self.clean_power
+        return self.income
+
+    def rate_service(self, mark):
+        new_sum_of_rating = \
+            self.average_rating * self.count_of_ratings + mark
+        self.average_rating = \
+            round(
+                new_sum_of_rating / (self.count_of_ratings + 1),
+                1)
```
DmytroSvirsa, line 44: split to different variables to make it more readable

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def calculate_washing_price(self, car):
+        price = car.comfort_class * \
+            (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car):
+        self.income += self.calculate_washing_price(car)
+        car.clean_mark = self.clean_power
+        return self.income
```
DmytroSvirsa, line 36: you have `return` but don't use it, remove statement

DmytroSvirsa (changes requested):

app/main.py
musakovskiy: New Version

PR #161, opened 2022-09-13
Solution

wQuelS:


app/main.py
```diff
@@ -1,8 +1,39 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_rate = self.clean_power - car.clean_mark
+        rate = car.comfort_class * self.average_rating
+        price = round(rate * clean_rate / self.distance_from_city_center, 1)
+        return price
+
+    def wash_single_car(self, car: Car):
```
Nattalli, line 32: don't forget to add type annotation for **every** function. Because 2/4 functions have a type annotation (the type which returns a function) and 2 haven't. Remember: if the function returns nothing then it has a type None.

Nattalli (changes requested):

Nattalli (approved):
Good!

PR #166, opened 2022-09-14
Solution

StasOkhrym:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car):
+        clean_requirement = self.clean_power - car.clean_mark
+        rating_ratio = self.average_rating / self.distance_from_city_center
+        total_price = clean_requirement * car.comfort_class * rating_ratio
+        return round(total_price, 1)
+
+    def rate_service(self, new_rate=int):
```
Nattalli, line 43: why do you use a type annotation like this `new_rate=int`?

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car):
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars=list) -> float:
```
Nattalli, line 28: and here too, fix your type annotation

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car):
+        clean_requirement = self.clean_power - car.clean_mark
+        rating_ratio = self.average_rating / self.distance_from_city_center
+        total_price = clean_requirement * car.comfort_class * rating_ratio
+        return round(total_price, 1)
+
+    def rate_service(self, new_rate: int):
```
Nattalli, line 43: and the last change you have to do - add type annotation for functions. If the function returns nothing than the type will be `None`

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list):
```
Nattalli, line 28: and for this functions too

Nattalli (changes requested):

Nattalli (approved):
Finally! 🔥

PR #176, opened 2022-09-16
Solution

Yaroslav-Stetsko:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class, clean_mark, brand):
```
Nattalli, line 2: add type annotation

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
```
Nattalli, line 9: add type annotation

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                 distance_from_city_center,
+                 clean_power,
+                 average_rating,
+                 count_of_ratings):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list):
```
Nattalli, line 19: add type annotation

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        income = 0
+        for car in car_list:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car):
```
Nattalli, line 27: add type annotation

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car):
```
Nattalli, line 32: add type annotation

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rating):
```
Nattalli, line 35: add type annotation

Nattalli (changes requested):

app/main.py
Yaroslav-Stetsko: Fixed

app/main.py
Yaroslav-Stetsko: Fixed

Yaroslav-Stetsko: Fixed

app/main.py
Yaroslav-Stetsko: Fixed

app/main.py
Yaroslav-Stetsko: Fixed

app/main.py
Yaroslav-Stetsko: Fixed

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car):
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car):
+        car.clean_mark = self.clean_power
```
Y-Havryliv, line 33: You need to check if `clean_power` is greater than `clean_mark` here as well because in case you run this method separately from others it will not work correctly

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv: Also, add annotations for return values:
```
```

Yaroslav-Stetsko: Done 
Done

Y-Havryliv (approved):
Well done!

PR #181, opened 2022-09-17
Solution

AndriyKy:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list):
+        cost = []
+        for car_instance in cars:
+
+            # Checking for the possibility of washing
```
IvanRamyk, line 22: Please, remove the redundant comment

Your variable names are good enough so you don't need comments.

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self, distance_from_city_center, clean_power,
+        average_rating, count_of_ratings
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list):
+        cost = []
```
IvanRamyk, line 19: It's better to use an integer variable to calculate the total cost
```suggestion
        cost = 0
```

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list):
+        cost = 0
+        for car_instance in cars:
+
+            car_was_washed = self.wash_single_car(car_instance)
+
+            # Return the amount for the service, otherwise False
```
IvanRamyk, line 24: this comment is redundant too

IvanRamyk (changes requested):

app/main.py
AndriyKy: And what's wrong with tests?

IvanRamyk (approved):
Good job!

PR #182, opened 2022-09-17
Added 2 classes with methods for car wash station

alexzubkoff:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+            if car.clean_mark < self.clean_power:
+                car_washed_income.append(self.calculate_washing_price(car))
+                self.wash_single_car(car)
+        return sum(car_washed_income)
+
+    def calculate_washing_price(self, car_inst: Car) -> float:
+        income = car_inst.comfort_class * \
+            (self.clean_power - car_inst.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
+        return round(income, 2)
```
Y-Havryliv, line 33: It should return a number rounded to 1 decimal.

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+
+    def wash_single_car(self, car_inst: Car) -> None:
+        car_inst.clean_mark = self.clean_power
```
Y-Havryliv, line 36: You need to check if `clean_power` is greater than `clean_mark` here as well because in case you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                self.wash_single_car(car)
+        return sum(car_washed_income)
+
+    def calculate_washing_price(self, car_inst: Car) -> float:
+        income = car_inst.comfort_class * \
+            (self.clean_power - car_inst.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
+        return round(income, 2)
+
+    def wash_single_car(self, car_inst: Car) -> None:
```
Y-Havryliv, line 35: Also, `car_inst` is a bad variable name, better to use `car_instance`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #183, opened 2022-09-17
Solution

VitaliiPiaskivskyi:


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E203, E266, W503
+ignore = E203, E266, W503, E501
```
Y-Havryliv, line 2: please, delete `E501`, it's not the solution you need)

app/main.py
```diff
@@ -1,8 +1,36 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int, average_rating: float, count_of_ratings: int):
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list):
```
Y-Havryliv, line 16: Also, add annotations for return values as well:
```
```

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        res = 0
```
Y-Havryliv, line 17: `res` is not a good variable name

Y-Havryliv (changes requested):

.flake8
VitaliiPiaskivskyi: [@Igor Kosyanenko](https://mateacademy.slack.com/team/U03Q39WJDBN)
 ні, краще додай в .flake8 помилку W503 в список помилок які мають ігноруватись

PR #185, opened 2022-09-18
'Solution'

rostyak:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list):
+        income = 0
+        for car in cars:
+            if self.clean_power >= car.clean_mark:
+                income += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
```
Y-Havryliv, line 21: you have a method for washing a car, use it

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        return income
+
+    def calculate_washing_price(self, car):
+        price = car.comfort_class * (self.clean_power - car.clean_mark) \
+            * self.average_rating / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car):
+        if self.clean_power >= car.clean_mark:
+            price = self.calculate_washing_price(car)
```
Y-Havryliv, line 31: you don't need to calculate washing price in here

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+
+    def serve_cars(self, cars: list):
+        income = 0
+        for car in cars:
+            if self.clean_power >= car.clean_mark:
+                income += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
+        return income
+
+    def calculate_washing_price(self, car):
```
Y-Havryliv, line 24: if you started using annotations, please use it everywhere

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv: Also, add annotations everywhere:
```
```

Y-Havryliv (approved):
Well done!

PR #187, opened 2022-09-18
Solution

matechris:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+    def calculate_washing_price(self, car):
+        distance = self.distance_from_city_center
+        rating = self.average_rating
+        difference = self.clean_power - car.clean_mark
+        price = car.comfort_class * difference * rating / distance
+
+        return round(price, 1)
+
+    def wash_single_car(self, car):
+        car.clean_mark = self.clean_power
```
Y-Havryliv, line 38: You need to check if `clean_power` is greater than `clean_mark` here as well because in case you run this method separately from others it will not work correctly

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv: Also, add annotations everywhere:
```
```

PR #190, opened 2022-09-18
'Solution'

serviza:


Y-Havryliv (changes requested):
tests haven't passed, fix please

serviza:
> Tests haven't passed, fix please. Also, add annotations everywhere.

Ok! Now it looks like fixed)

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+    def serve_cars(self, cars: list):
+        cost = 0
+        # calculate total cost of washing all cars
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                cost += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(cost, 1)
+
+    def calculate_washing_price(self, car):
```
Y-Havryliv, line 26: Add annotations everywhere, and for return values as well, for example:
```
```

Y-Havryliv (changes requested):

app/main.py
serviza: > Add annotations everywhere, and for return values as well, for example:
> 
> ```
> def rate_service(self, rating: float) -> None:
> ```

Done.

Y-Havryliv (approved):
Well done!

PR #191, opened 2022-09-18
solution

FranchukS:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+
+    def calculate_washing_price(self, car: Car):
+        return round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center,
+            1)
+
+    def wash_single_car(self, car: Car):
+        car.clean_mark = self.clean_power
```
TPolina, line 36: You need to check if `clean_power` is greater than `clean_mark` here as well because in case you run this method separately from others it will not work correctly

TPolina (changes requested):

app/main.py
FranchukS: Fixed, also added annotation for return in methods

TPolina (approved):
Well done!

PR #195, opened 2022-09-19
Solution

HStakhiv:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list):
```
FranchukS, line 21: Add annotation for return value in all methods:
def serve_cars(self, cars: list) -> 'value':

FranchukS (changes requested):

app/main.py
HStakhiv: Done!

HStakhiv:
Is everything good now?

FranchukS (approved):
As for me, looks good)

Y-Havryliv (approved):
GJ!

PR #197, opened 2022-09-19
Solution to CarWashStation

pavlo-kapitanets:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                 / self.distance_from_city_center)
+        round_price = round(price, 1)
+        return round_price
+
+    def wash_single_car(self, car: Car):
+        car.clean_mark = self.clean_power
+        return car.clean_mark
+
+    def serve_cars(self, list_of_cars: list):
+        sorted_list_of_cars = [car for car in list_of_cars
```
mltdll, line 32: This variable should probably be named `filtered_list_of_cars`, since **"sorted"** usually means **"ordered"**.

mltdll (approved):

app/main.py
```diff
@@ -29,10 +29,10 @@ def wash_single_car(self, car: Car):
         return car.clean_mark
 
     def serve_cars(self, list_of_cars: list):
-        sorted_list_of_cars = [car for car in list_of_cars
+        filtered_list_of_cars = [car for car in list_of_cars
                                if car.clean_mark < self.clean_power]
```
mltdll, line 33: You also need to shift this line by a few spaces so that flake8 test passes. :)

pavlo-kapitanets: All done✓

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        price = (car.comfort_class
+                 * (self.clean_power - car.clean_mark)
+                 * self.average_rating
+                 / self.distance_from_city_center)
+        round_price = round(price, 1)
+        return round_price
+
+    def wash_single_car(self, car: Car):
+        car.clean_mark = self.clean_power
+        return car.clean_mark
```
TPolina, line 29: You need to check if `clean_power` is greater than `clean_mark` here as well because in case you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+    def serve_cars(self, list_of_cars: list):
+        filtered_list_of_cars = [car for car in list_of_cars
+                                 if car.clean_mark < self.clean_power]
```
TPolina, line 33: What do you need an additional list for? You can just add a check inside for loop

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        list_of_prices = []
```
TPolina, line 34: And what do you need this list for? Can't you just create income=0 and then add the value of washing_price each time

TPolina (changes requested):

TPolina (approved):
Well done!

PR #202, opened 2022-09-19
'Solution'

Ilyakson:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
-    pass
+
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
```
ketstap162, line 17: i think all variables should be in new line in __init__( ... )

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int):
```
ketstap162, line 12: I think all arguments in __init__( ... ) should be in new lines

ketstap162: "self" and "distance_from_city_center" too

ketstap162 (approved):

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, car: Car):
+        price = car.comfort_class * \
+            (self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
```
TPolina, line 33: Please, don't use `* \ `, you can add W504 flake8 rule to ignored ones or create a few temporal variables

TPolina (changes requested):

TPolina (approved):

PR #204, opened 2022-09-19
Solution

dmytro-petliovanyi:


mariia-movchan (approved):
Good job!
But in the future try to use type annotation :)

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center,
```
Y-Havryliv, line 11: Add annotations everywhere, and for return values as well, for example:
```
```

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+            average_rating,
+            count_of_ratings
+    ):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car):
+        work = (self.clean_power - car.clean_mark)
```
Y-Havryliv, line 22: you don't need parentheses here

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv, line 11: the annotation is still missing

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #206, opened 2022-09-19
Develop2

VitaliiPiaskivskyi:


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E203, E266, W503
+ignore = E203, E266, W503, E501
```
Y-Havryliv, line 2: you still need to delete `E501`

app/main.py
```diff
@@ -1,8 +1,36 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int, average_rating: float, count_of_ratings: int):
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list) -> int:
```
Y-Havryliv, line 16: it does not return `int`

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        result = 0
+        for car in list_of_cars:
+            if car.clean_mark < self.clean_power:
+                result += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(result, 1)
+
+    def calculate_washing_price(self, test_car: Car) -> int:
```
Y-Havryliv, line 24: same here

Y-Havryliv (changes requested):

.flake8
VitaliiPiaskivskyi: It is possible to bypass the 501 error through additional variables. Discussed on QNA. In this particular case, it is better to add an exception inside the flake8

app/main.py
VitaliiPiaskivskyi: It is possible to bypass the 501 error through additional variables. Discussed on QNA. In this particular case, it is better to add an exception inside the flake8

Y-Havryliv (approved):

PR #216, opened 2022-09-20
Solution

meloman-13:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                 * self.average_rating
+                 ) / self.distance_from_city_center
+
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int):
```
kateryna3k, line 47: add here type annotation as you did before

kateryna3k (changes requested):

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
```
Y-Havryliv, line 26: this method does not return `int`

Y-Havryliv (changes requested):

PR #246, opened 2022-10-20
Solution

tinabushuieva:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+
+    def calculate_washing_price(self, car: object) -> Union:
+        difference = self.clean_power - car.clean_mark
+        price = round(((car.comfort_class
+                        * difference * self.average_rating)
+                       / self.distance_from_city_center), 1)
+        return price, difference
+
+    def wash_single_car(self, car: object) -> str:
+        print(f"{car.brand} was washed")
```
DmytroSvirsa, line 44: this method must check if car can be wased by station and wash it if that is true

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                        * difference * self.average_rating)
+                       / self.distance_from_city_center), 1)
+        return price, difference
+
+    def wash_single_car(self, car: object) -> str:
+        if car.clean_mark <= self.clean_power:
+            return True
+        else:
+            False
+
```
Abnormaltype, line 48: this method should return nothing

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+    def rate_service(self, rate: float) -> float:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
+                                     * (self.count_of_ratings - 1) + rate)
+                                    / self.count_of_ratings, 1)
+        return self.average_rating
```
Abnormaltype, line 54: the same, should return nothing

Abnormaltype (changes requested):
please, fix type annotation, and put the correct

app/main.py
tinabushuieva: I don't get it if "this method must check if car can be washed by station and wash it if that is true"

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        income = 0
+        for car in cars:
+            wash = self.wash_single_car(car)
+            if wash:
+                price, difference = self.calculate_washing_price(car)
+                income += price
+                car.clean_mark += difference
+        return income
+
+    def calculate_washing_price(self, car: object) -> Union:
```
Abnormaltype, line 34: put the correct type annotation

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        difference = self.clean_power - car.clean_mark
+        price = round(((car.comfort_class
+                        * difference * self.average_rating)
+                       / self.distance_from_city_center), 1)
+        return price, difference
+
+    def wash_single_car(self, car: object) -> bool:
```
Abnormaltype, line 41: put the correct type annotation

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Abnormaltype, line 21:
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        price = round(((car.comfort_class
+                        * difference * self.average_rating)
+                       / self.distance_from_city_center), 1)
+        return price, difference
+
+    def wash_single_car(self, car: Car) -> bool:
+        if car.clean_mark <= self.clean_power:
+            return True
+
+    def rate_service(self, rate: float) -> None:
```
Abnormaltype, line 42:
```suggestion
    def rate_service(self, rate: Union[int, float]) -> None:
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #262, opened 2022-11-03
Develop

ochernyi:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        total_price = 0
+        for car in cars_list:
+            if car.clean_mark < self.clean_power:
+                income_from_washing_single_car = \
+                    self.calculate_washing_price(car)
+                total_price += income_from_washing_single_car
+                self.wash_single_car(car)
+        return total_price
+
+    def calculate_washing_price(self, instance: Car) -> float:
```
IvanRamyk, line 29:
```suggestion
    def calculate_washing_price(self, car: Car) -> float:
```

`instance` isn't a descriptive name

IvanRamyk (changes requested):

PR #270, opened 2022-11-08
Develop

phaishuk:
done

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,54 +1,8 @@
 class Car:
-    def __init__(
-            self,
-            comfort_class: int,
-            clean_mark: int,
-            brand: str
-    ) -> None:
-        self.comfort_class = comfort_class
-        self.clean_mark = clean_mark
-        self.brand = brand
+    # write your code here
+    pass
 
 
 class CarWashStation:
-    def __init__(
-            self,
-            distance_from_city_center: float,
-            clean_power: int,
-            average_rating: float,
-            count_of_ratings: int
-    ) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        return final_price
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (
+                self.clean_power - car.clean_mark
+            ) * (
+                self.average_rating / self.distance_from_city_center
+            ), 1
+        )
```
DmytroSvirsa, line 43: If you have some long math, you can split it onto additional variables, or break after binary operations (not before - it cause the W504 errors) read checklist

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rate: int) -> None:
+        self.average_rating = round(
+            (
+                (self.average_rating * self.count_of_ratings) + new_rate
+            ) / (self.count_of_ratings + 1), 1
+        )
+        self.count_of_ratings += 1
```
DmytroSvirsa, line 55: the same

DmytroSvirsa (changes requested):

app/main.py
phaishuk: solved

app/main.py
phaishuk: solved

DmytroSvirsa (approved):
GJ

PR #271, opened 2022-11-09
Solution

andriy-md:
I had some style problems: flake8 indicated that my lines were too long.
So I've tried to just break a long line into many lines by pressing "enter" (as the internet suggested doing).

Hope, this is a suitable solution.

Thanks!


DmytroSvirsa (approved):
that is ok, but read checklist it can save your time

PR #279, opened 2022-11-16
py_car_wash_station

Volynets83:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> int or str:
```
Serhii-Leonenko, line 3: Are you sure that this method returns int or str?) 
The same in other similar cases

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: float, average_rating: float,
+                 count_of_ratings: float) -> int or float:
```
Serhii-Leonenko, line 12: Fix type annotations, `clean_power` and `count_of_rating` aren't ` float`.
And the same as previous comment

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> int or float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, cars: Car) -> int or float:
```
Serhii-Leonenko, line 26: This method returns only float, because it returns values rounded to 1 decimal, the same in `serve_cars.` And why `cars`, this method takes only a single car?

Serhii-Leonenko (changes requested):

Abnormaltype (approved):
Well done!

PR #285, opened 2022-11-20
Solution

Yelinskyi:


app/main.py
```diff
@@ -1,8 +1,56 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> Car:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+        return car
```
Abnormaltype, line 43: this method should return nothing

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+            (self.average_rating * self.count_of_ratings + mark)
+            / (self.count_of_ratings + 1),
+            1
+        )
+        self.count_of_ratings += 1
+
+
+# Readmi file has a bit mistake. It's confusing
+# income == 41.7
+# but must be 41.6
```
Abnormaltype, line 56: thx, but you can write this in a comment to the PR) remove comments)

Abnormaltype (changes requested):

app/main.py
Yelinskyi: Sure.
But I don't have ather way to communicate.

Abnormaltype (approved):

PR #311, opened 2022-12-13
Develop

TarasSavchyn:


app/main.py
```diff
@@ -1,8 +1,53 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str)\
```
katevasilenko, line 5: Just do this in way you made it in `CarWashStation` class so it will look a bit better

katevasilenko (approved):
Looks good to me

PR #316, opened 2022-12-17
Remade

timyr4iq:
made changes

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        return round((car.comfort_class
+                      * (self.clean_power - car.clean_mark)
+                      * self.average_rating
+                      / self.distance_from_city_center), 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = \
```
katevasilenko, line 40: It’s picky, but you can avoid to use`\ ` in code: `self.average_rating = round(` and than go to next line

katevasilenko (approved):
Good solution, I just would prefer one small change.

PR #327, opened 2022-12-28
Solution py-car-wash-station

avkpol:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+            clean_power: int, average_rating: float,
+            count_of_ratings: int,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        cws = CarWashStation
```
katevasilenko, line 22: You don’t need this

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += cws.calculate_washing_price(self, car)
```
katevasilenko, line 26: You already have `self`, just use it instead of `cws` And remove from arguments

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                car.clean_mark = cws.wash_single_car(self, car)
```
katevasilenko, line 27: The same here

katevasilenko (changes requested):

IvanRamyk (approved):

olena-hul (changes requested):

PR #330, opened 2023-01-02
'Solution'

Serhii-Bezimov:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> (float, int):
```
Serhii-Leonenko, line 18:
```suggestion
    def serve_cars(self, car_list: list[Car]) -> float:
```

It returns only float because result number is rounded to 1 decimal

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        sum_calculate = 0
+        for car in car_list:
+            if self.clean_power > car.clean_mark:
```
Serhii-Leonenko, line 21: You have the same check in the wash_single_car() method, so use just it.

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+                * self.average_rating / self.distance_from_city_center), 1
+        )
+
+    def wash_single_car(self, car: Car) -> int:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+            return car.clean_mark
+
+    def rate_service(self, rating: int) -> (float, int):
```
Serhii-Leonenko, line 37:
```suggestion
    def rate_service(self, rating: int) -> float:
```

Serhii-Leonenko (changes requested):

PR #331, opened 2023-01-03
Solution

Rename925:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
olena-hul, line 3: Put each argument into separate line. It will look more readable. If all arguments don't fit into 1 line, it's better to place them in separate lines

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
olena-hul, line 11: The same comment about arguments applies here

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
```
olena-hul, line 17: Specify type of elements in list in the annotation (it will be list[Car]). And return annotation is incorrect. Do we round integers?

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        cost = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                cost += self.calculate_washing_price(car)
+            self.wash_single_car(car)
```
olena-hul, line 22: This method call should be inside if

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        return round(cost, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
```
olena-hul, line 31: You have the similar check inside `serve_cars`

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        cost = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                cost += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return round(cost, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
```
olena-hul, line 27: Format the maths better. Use more lines and `()`

olena-hul (changes requested):

katevasilenko (approved):
Good solution

PR #347, opened 2023-01-17
Done

numerodiciannove:


app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
olena-hul, line 4: Place each argument on separate line

olena-hul (changes requested):

app/main.py
numerodiciannove: updated

olena-hul (approved):

PR #350, opened 2023-01-20
Solution

Roman28101:


app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
olena-hul, line 4:
```suggestion
    def __init__(
        self, 
        comfort_class: int,
        clean_mark: int,
        brand: str
    ) -> None:
```

olena-hul: Follow this example and fix in all places

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        money_list = []
```
olena-hul, line 20: no need to declare list here. Just `income = 0` will be enough

app/main.py
```diff
@@ -1,8 +1,42 @@
...
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
olena-hul, line 19:
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        money_list = []
+        for car in cars:
+            if car.clean_mark <= self.clean_power:
+                money_list.append(
+                    CarWashStation.calculate_washing_price(self, car=car))
```
olena-hul, line 24:
```suggestion
                    income += self.calculate_washing_price(car)
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                CarWashStation.wash_single_car(self, car=car)
```
olena-hul, line 25: Don't call instance methods like this. Use `self.<method_name>`

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        return round(sum(money_list), 1)
+
+    def calculate_washing_price(self, car: str) -> float:
```
olena-hul, line 28: car is not str

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        income = (car.comfort_class * (self.clean_power - car.clean_mark)
+                  * self.average_rating) / self.distance_from_city_center
+        return income
+
+    def wash_single_car(self, car: str) -> float:
+        car.clean_mark = self.clean_power
+        return car.clean_mark
```
olena-hul, line 35: No need to return here

olena-hul (changes requested):

app/main.py
olena-hul: And car is not str

Vasyl-Smutok (approved):
Good Job 👍

PR #351, opened 2023-01-20
'Solution'

Anyta17:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        suma = 0.0
+        for car in cars:
+            if self.clean_power >= car.clean_mark:
+                suma += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return round(suma, 1)
```
YuriiKindrat, line 25: do u need to use round here? if u use round in calculate_washing_price func

YuriiKindrat (changes requested):

PR #353, opened 2023-01-21
Solution

bohdan-yatsyna:


app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
```
ArsenPidhoretskyi, line 6: Add validation for this field:

> comfort_class - comfort class of a car, from 1 to 7

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.clean_mark = clean_mark
```
ArsenPidhoretskyi, line 7: Add validation for this field:

> clean_mark - car cleanness mark, from very dirty - 1 to absolutely clean - 10

app/main.py
```diff
@@ -1,8 +1,44 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
```
ArsenPidhoretskyi, line 17: Add validation for this field:

> distance_from_city_center - how far station from the city center, from 1.0 to 10.0

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
```
ArsenPidhoretskyi, line 19: Add validation for this field:

> average_rating - average rating of the station, from 1.0 to 5.0, rounded to 1 decimal

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        cleaning_difficulty = self.clean_power - car.clean_mark
+        single_cleaning_price = (car.comfort_class
+                                 * cleaning_difficulty
+                                 * self.average_rating
+                                 / self.distance_from_city_center)
+        return round(single_cleaning_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
ArsenPidhoretskyi, line 39: You need also to add check for:

>  if wash_station.clean_power is greater than car.clean_mark;

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cleaning_difficulty = self.clean_power - car.clean_mark
+        single_cleaning_price = (car.comfort_class
+                                 * cleaning_difficulty
+                                 * self.average_rating
+                                 / self.distance_from_city_center)
```
ArsenPidhoretskyi, line 35:
```suggestion
        single_cleaning_price = (
            car.comfort_class * cleaning_difficulty
            * self.average_rating / self.distance_from_city_center
        )
```

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        return round(single_cleaning_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: (int, float)) -> None:
```
ArsenPidhoretskyi, line 41: Use `Union[int, float]` from typing module or just `int | float`

ArsenPidhoretskyi (changes requested):

app/main.py
olena-hul: I think those validation are unnecessary and just make code more complicated. Of course, it is a good practice to perform validations, but this task doesn't focus on it and we don't require from students to make validations in this task

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        cleaning_difficulty = self.clean_power - car.clean_mark
+        single_cleaning_price = (
+            car.comfort_class * cleaning_difficulty
+            * self.average_rating / self.distance_from_city_center
+        )
+        return round(single_cleaning_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
olena-hul, line 43: It is not necessary, because we only call this method from serve_cars method and the check is already included

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
```
olena-hul, line 32: The rounding is also redundant here, because you already perform rounding inside calculate_washing_price

olena-hul (changes requested):

app/main.py
bohdan-yatsyna: Funny the fact that exactly this change was asked in the previous changes request :-)

app/main.py
bohdan-yatsyna: Thanks for your remark.

bohdan-yatsyna:
My apologies for the confusion about the review request, I am not fully understand how it should works + bad connection did not lead me favor.

Vasyl-Smutok (approved):
Well Done 👍

PR #360, opened 2023-01-29
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
olena-hul, line 4: Place each argument in 1 line

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round((self.average_rating
+                                     * self.count_of_ratings
+                                     + rate)
+                                    / (self.count_of_ratings + 1),
+                                    1)
+        self.count_of_ratings += 1
+
+    def serve_cars(self, cars: list) -> float:
```
olena-hul, line 38:
```suggestion
    def serve_cars(self, cars: list[Car[) -> float:
```

app/main.py
olena-hul: Fix in all places

olena-hul (changes requested):

app/main.py
VladyslavKaminskyi: If I place arguments in one line. Flake8 : E501 line too long

PR #364, opened 2023-01-30
Add final solution

ZaikaDaria:


app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
katevasilenko, line 3:
```suggestion
    def __init__(
             self,
             comfort_class: int,
             clean_mark: int,
             brand: str
     ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
katevasilenko, line 12: Please put all parameters to a new line as I suggested you above

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+    def serve_cars(self, cars: list[Car]) -> float:
+        total_sum = 0.0
+        for car in cars:
+            total_sum += self.wash_single_car(car)
+        return total_sum
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = car.comfort_class * (
+            self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
```
katevasilenko, line 27: The line should start with operator. In that way you will avoid to use \

katevasilenko (changes requested):

Abnormaltype (approved):
Good job!

PR #369, opened 2023-02-02
add solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,53 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+
+        if len(cars) == 1:
+            income += self.calculate_washing_price(cars[0])
+            self.wash_single_car(cars[0])
+
+        else:
+            for car in cars:
+                if car.clean_mark < self.clean_power:
+                    income += self.calculate_washing_price(car)
+                    car.clean_mark = self.clean_power
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        diff = self.clean_power - car.clean_mark
+        if diff < 0:
+            return 0
+        calc = round((car.comfort_class * diff * self.average_rating
+                      / self.distance_from_city_center), 1)
+        return calc
+
+    def rate_service(self, mark: int) -> None:
+        self.average_rating = round((self.average_rating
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,53 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
```
YuriiKindrat, line 17: fix identation here, please

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+
+        if len(cars) == 1:
+            income += self.calculate_washing_price(cars[0])
+            self.wash_single_car(cars[0])
+
+        else:
```
YuriiKindrat, line 30: why do u need this condition ? if list will have only 1 element, u could also use for loop to iter through it

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+            for car in cars:
+                if car.clean_mark < self.clean_power:
+                    income += self.calculate_washing_price(car)
+                    car.clean_mark = self.clean_power
```
YuriiKindrat, line 34: u can use wash_single_car method here

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        diff = self.clean_power - car.clean_mark
+        if diff < 0:
+            return 0
```
YuriiKindrat, line 40: no need to make this check here, car.clean_mark always will be less then self.clean_power, u check this in serve_cars method using if condition

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        calc = round((car.comfort_class * diff * self.average_rating
```
YuriiKindrat, line 41: use full names for variables

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        return calc
+
+    def rate_service(self, mark: int) -> None:
+        self.average_rating = round((self.average_rating
+                                     * self.count_of_ratings + mark)
+                                    / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
YuriiKindrat, line 52: no need to make this condition here, u already check this in serve_cars method

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: I fixed my code according u messages

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        calc = round((car.comfort_class
```
YuriiKindrat, line 32: use full names for variables, "calc" is not full name :)
it is better to call this variable f.e. washing_price or smth else

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: done

Serhii-Leonenko (approved):
Well done!

manjustice:
GJ!

manjustice (commented):
GJ!

manjustice (approved):

PR #373, opened 2023-02-03
Solution

bihunva:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        single_income = round(
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * self.average_rating)
+            / self.distance_from_city_center, 1)
+
+        return single_income
+
+    def wash_single_car(self, car: Car) -> bool:
+        return self.clean_power > car.clean_mark
```
YuriiKindrat, line 45: image
fix this, according to requirements

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                income += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        single_income = round(
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * self.average_rating)
+            / self.distance_from_city_center, 1)
```
YuriiKindrat, line 40: fix identation here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        single_income = round(
+            (car.comfort_class
+             * (self.clean_power
```
wQuelS, line 38: Can you decompose your formula into smaller chunks with the help of additional variables? It will not only make things more clear, but will increase readability of your code

wQuelS (changes requested):
Another advice is to use third-party libraries, like black, to format your .py files automatically
Documentation: https://pypi.org/project/black/

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        difference = self.clean_power - car.clean_mark
+        single_income = round(
+            car.comfort_class
+            * difference
```
wQuelS, line 39: It still looks bad, and it is not easy to read and understand what is going on in your formula. I bet you can divide your formula even more:)

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        single_income = round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
```
wQuelS, line 41: please, separate this huge formula into smaller chunks, using additional variables. Such an approach will guarantee better readability and your code will be more maintainable.

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+        comfort_class = car.comfort_class
+
+        single_income = round(
+            comfort_class * difference * ave_rating / distance
+            , 1)
+
+        return single_income
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Serhii-Leonenko, line 51: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #374, opened 2023-02-03
Solution

ZabFTFT:


app/main.py
```diff
@@ -1,8 +1,48 @@
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
YuriiKindrat, line 5: fix identation here, according to the best practicies

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center : int, clean_power: int,
+                 average_rating: int, count_of_ratings: int) -> None:
```
YuriiKindrat, line 14: the same

YuriiKindrat (changes requested):

app/main.py
ZabFTFT: Read and fixed.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        diff_clean_power = self.clean_power - car.clean_mark
+        wash_price_wo_city_center = (car.comfort_class
+                                     * diff_clean_power
+                                     * self.average_rating)
+        washing_price_with_cc = (wash_price_wo_city_center
+                                 / self.distance_from_city_center)
```
wQuelS, line 38: Your line break look not so good. Can you find out way to get rid of them?

wQuelS (changes requested):

app/main.py
ZabFTFT: I made some additional method to solve this.Is it correct form to fix this issuer?

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

ZabFTFT:
> Good job!

Thank you very much for your effort!

PR #375, opened 2023-02-03
Solution

kuznetsb:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
Serhii-Leonenko, line 3: Separate parameters for better readability:

```suggestion
    def __init__(self,
                 comfort_class: int,
                 clean_mark: int, 
                 brand: str) -> None:
```

Also in all other similar cases

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Serhii-Leonenko, line 19: You can give more descriptive type annotations:

```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
+        return round(cost, 1)
```
Serhii-Leonenko, line 30: Read the checklist


image

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Serhii-Leonenko, line 33: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -25,12 +28,15 @@ def serve_cars(self, cars: list) -> float:
         return round(income, 1)
 
     def calculate_washing_price(self, car: Car) -> float:
-        cost = car.comfort_class * (self.clean_power - car.clean_mark) * \
-            self.average_rating / self.distance_from_city_center
+        cost = car.comfort_class\
+            * (self.clean_power - car.clean_mark)\
+            * self.average_rating\
+            / self.distance_from_city_center
         return round(cost, 1)
```
Zenionoi, line 35: In general, you don't need to use the `\` symbol, if you take the expression in parentheses `()`, it was in the mentor's remark. You can also avoid creating an additional variable and return the result at once

Zenionoi (changes requested):

Zenionoi (approved):

Serhii-Leonenko (approved):
Well done!

PR #376, opened 2023-02-03
Solution

WDemidenko:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            diff = (self.clean_power - car.clean_mark)
```
NEkropo1, line 29: This code is ok for this case, but, I think, in case if we going to extend services(for DRY principle), slightly better to implement single function, which will check if station could clean car.

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            return round(
+                (
+                    (
+                        car.comfort_class * diff * self.average_rating
+                    ) / self.distance_from_city_center
+                ), 1
+            )
+        return 0
```
NEkropo1, line 37: Don't agree 0 is float(small tip u can return 0.0 or use Union from typing)
But more important - u return 0 which is never used in code, because only implementation of code in your code is when CarWashStation.clean_power > Car.clean_mark, so this return is redundant

NEkropo1 (changes requested):

NEkropo1 (approved):
Good job!

Serhii-Leonenko (approved):
Good job!

PR #377, opened 2023-02-04
'Solution'

alexlega:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_rating
+        self.to_wash = 0
+
+    def serve_cars(self, list_of_cars: list) -> float:
+        return sum([self.calculate_washing_price(car) for car in list_of_cars
+                    if self.wash_single_car(car)])
+
+    def calculate_washing_price(self, other: typing.Any) -> float:
+        return round(other.comfort_class * (
```
SergiiMachulin, line 27: Fix indentation here

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+
+    def wash_single_car(self, other: typing.Any) -> bool:
+        if other.clean_mark < self.clean_power:
+            self.to_wash = other.clean_mark
+            other.clean_mark = self.clean_power
+
+            return True
+
+    def rate_service(self, single_rating: int) -> None:
+        self.average_rating = round(
```
SergiiMachulin, line 39: Fix indentation here

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_rating: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_rating
+        self.to_wash = 0
+
+    def serve_cars(self, list_of_cars: list) -> float:
```
SergiiMachulin, line 22: Add annotation of class inside list

app/main.py
```diff
@@ -1,8 +1,42 @@
...
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_rating: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_rating
+        self.to_wash = 0
```
SergiiMachulin, line 20: I think this attribute is unnecessary

SergiiMachulin (commented):
Please check my comments

app/main.py
alexlega: Without this attribute, my program doesn't work

app/main.py
```diff
@@ -19,24 +16,25 @@ def __init__(self, distance_from_city_center: float, clean_power: int,
...
         return sum([self.calculate_washing_price(car) for car in list_of_cars
                     if self.wash_single_car(car)])
 
-    def calculate_washing_price(self, other: typing.Any) -> float:
-        return round(other.comfort_class * (
-            self.clean_power - self.to_wash)
-            * self.average_rating / self.distance_from_city_center, 1)
+    def calculate_washing_price(self, other: Car) -> float:
+        return round(other.comfort_class
+                     * (self.clean_power - self.to_wash)
```
ada-krav, line 25: Per our assignment "car's comfort class * difference between wash station's clean power and **_car's clean mark_* * * car wash station rating / car wash station distance to the center of the city", so self.to_wash -> other.clean_mark . 
While it probably will work withing **serve cars**, **calculate_washing_price** separately will give the wrong price.

Also, maybe other is not the best choice for an object name. Maybe just "car"?

app/main.py
```diff
@@ -1,8 +1,40 @@
...
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_rating: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_rating
+        self.to_wash = 0
```
ada-krav, line 17: In README "Create class CarWashStation, its constructor takes and stores 4 arguments:" 
Adding "to_wash" makes constructor store 5 arguments.

ada-krav (changes requested):

ada-krav (approved):

app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
Serhii-Leonenko, line 3: Use indentation like this to increase the readability of your code. The same in other similar cases.


```suggestion
    def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str
) -> None:
```

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            price = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return price
+        else:
```
Serhii-Leonenko, line 34: Remove redundant else

Serhii-Leonenko (changes requested):

app/main.py
alexlega: But in this case I will have sum of integers and nonetypes's in serve_cars, when wash_single_car is called

alexlega: I understood my mistake and changed some logic in my methods. Now the program works correctly without "else: return 0"

YuriiKindrat (approved):

PR #378, opened 2023-02-04
Solution

NEkropo1:
added can_we_check_car for class 'CarWashStation' :
`help(CarWashStation.can_we_wash_car)`

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,60 @@
+from typing import Union
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int, clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+        self,
+        distance_from_city_center: Union[float, int],
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def can_we_wash_car(self, car: Car) -> bool:
+        """
+        returns True if station.clean_power > car.clean mark else False
+        """
+        return self.clean_power > car.clean_mark
+
+    def serve_cars(self, cars: list) -> float:
+        total_income = 0
+        for car in cars:
+            if self.can_we_wash_car(car):
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return total_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            (self.clean_power - car.clean_mark)
+            * car.comfort_class
+            * self.average_rating
+            / self.distance_from_city_center,
+            1,
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.can_we_wash_car(car):
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, mark: Union[float, int]) -> float:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + mark)
+            / (self.count_of_ratings + 1),
+            1,
+        )
+        self.count_of_ratings += 1
+        return self.average_rating


app/main.py
```diff
@@ -1,8 +1,60 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def can_we_wash_car(self, car: Car) -> bool:
+        """
+        returns True if station.clean_power > car.clean mark else False
+        """
+        return self.clean_power > car.clean_mark
```
SergiiMachulin, line 29: I think it' also must be True if self.clean_power = car.clean_mark

app/main.py
```diff
@@ -1,8 +1,60 @@
...
+
+    def serve_cars(self, cars: list) -> float:
```
SergiiMachulin, line 31: I think it's better to give more descriptive type annotation inside list

app/main.py
```diff
@@ -1,8 +1,60 @@
...
+            * self.average_rating
+            / self.distance_from_city_center,
+            1,
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.can_we_wash_car(car):
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, mark: Union[float, int]) -> float:
```
SergiiMachulin, line 53: According to README.md average_rating and count_of_ratings only should be changed. It's unnecessary return in this method.

SergiiMachulin (commented):
Please check my comments

app/main.py
NEkropo1: Since it won't afford in any way I disagree. Yup, making Generics is more descriptive, but, really, according to README.md, we only producing list of Car's and if we make this annotation it won't help, till we get logs, so it's just excessive
image

app/main.py
NEkropo1: CarWashingStation provides services for Cars for getting income. If we get clean_power == car.clean_mark, we get 0 income in a formula, while wasting resources on washing car. So I disagree on this.

app/main.py
NEkropo1: Yup totally agree on that. Since i don't know architecture of overall app, it could break tests, or some services.

app/main.py
```diff
@@ -1,8 +1,59 @@
+from typing import Union
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int, clean_mark: int, brand: str) -> None:
```
viktoria-rybenchuk, line 6: When you’re using a hanging indent, there must not be any arguments on the first line. The following example is not PEP 8 compliant

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def can_we_wash_car(self, car: Car) -> bool:
+        """
+        returns True if station.clean_power > car.clean mark else False
+        """
+        return self.clean_power > car.clean_mark
```
viktoria-rybenchuk, line 29: with this function you brake the DRY principle. You can solve this task without this method

viktoria-rybenchuk (changes requested):

app/main.py
NEkropo1: WET > DRY
KISS, YAGNI, DRY:
We implement function which is usable more than 1 times, and which makes code more readable + if we implement any future features, this method is crucial 'cause it's only parametr we compare to decide, either we make anything or not, so I totally disagree.

app/main.py
NEkropo1: It is PEP 8 compilant, cause self is just link on object, while others are parameters. Furthermore, here is just example from pep:
image
But it's just simple case from functions, while there are much more complicated cases, like in python modules:
image
The only thing I'm not totally sure is - either I need or not have all other parametres in hanging way, but black just reformated file this way, so all questions to https://pypi.org/project/black/

app/main.py
viktoria-rybenchuk: It makes sense if  you need to compare objects frequently, it might be more efficient to implement a special comparison function and use it in your code. My bad.

viktoria-rybenchuk (approved):

app/main.py
Serhii-Leonenko, line 6: It's better to use the same code style in all your code. So use indentation similar to the CarWashStation, also it has better readability

Serhii-Leonenko (changes requested):

app/main.py
NEkropo1: > It's better to use the same code style in all your code. So use indentation similar to the CarWashStation, also it has better readability

So, is it ok, if my indentations throughout code will be like this?
image

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,64 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: Union[float, int],
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int,
+    ) -> None:
```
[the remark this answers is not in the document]
NEkropo1: Yup, I somehow reworked other file while pushed this )

app/main.py
```diff
@@ -1,8 +1,64 @@
...
+        for car in cars:
+            if self.can_we_wash_car(car):
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return total_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            (self.clean_power - car.clean_mark)
```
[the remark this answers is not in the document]
NEkropo1: I was sure that formula is ok, like every step is readable and countable according to exact formula steps in readme...
I reworked it, making smaller parts, but it seems to me, that old one was more readable)

wQuelS (approved):
Good job

PR #381, opened 2023-02-05
solution

Yulia-Mala:


usrofgh:
You can also use same names of attributes and arguments in `__init__()`. But it isn't a PEP 8 rule and the code works. So everything is ok. Good job ;)

Yulia-Mala:
I try to use "in_arg" names in funcs for clear understanding which data came from outside. but in init it's hard to mix it, you are right)

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.distance_from_city_center = in_distance
+        self.clean_power = in_power
+        self.average_rating = in_rating
+        self.count_of_ratings = in_count
+
+    def calculate_washing_price(self, in_car: Car) -> float:
+        price = in_car.comfort_class                    \
+            * (self.clean_power - in_car.clean_mark)    \
+            * self.average_rating                       \
+            / self.distance_from_city_center
```
SergiiMachulin, line 26: According to the checklist to this task presence of \ sign is bad practice, please fix it

app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 in_class: int,
+                 in_mark: int,
+                 in_brand: str) -> None:
```
SergiiMachulin, line 5: [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation) specifically gives an example for such case, so perhaps the mainstream is going to adapt that style. 
Personally I advise to use separating each parameter into its own line, so that any future addition/deletion of parameter would give a clean ```git diff```
```
def __init__(
    self,
    in_class: int,
    in_mark: int,
    in_brand: str
) -> None:
```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 in_distance: float,
+                 in_power: int,
+                 in_rating: float,
+                 in_count: int) -> None:
```
SergiiMachulin, line 16: The same indentation advise as above

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def calculate_washing_price(self, in_car: Car) -> float:
+        price = in_car.comfort_class                    \
+            * (self.clean_power - in_car.clean_mark)    \
+            * self.average_rating                       \
+            / self.distance_from_city_center
+        price = round(price, 1)
+        return price
+
+    def wash_single_car(self, in_car: Car) -> None:
+        in_car.clean_mark = self.clean_power
```
SergiiMachulin, line 31: According to README.md for this method next condition exits: ```if wash_station.clean_power is greater than car.clean_mark```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def serve_cars(self, cars_to_wash: list[Car]) -> float:
+        income = 0
+        for car in cars_to_wash:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        income = round(income, 1)
```
SergiiMachulin, line 39: I think you can avoid this line with addition the same logic to return

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        return income
+
+    def rate_service(self, feedback: int) -> None:
+        new_rate = (self.average_rating * self.count_of_ratings + feedback) / \
```
SergiiMachulin, line 43: Please fix indentation according to the task checklist

SergiiMachulin (changes requested):

app/main.py
Yulia-Mala: ok

app/main.py
Yulia-Mala: if you mentioned PEP - please give a certain PEP index. For example: "pls check  PEP 123"

app/main.py
Yulia-Mala: ok

app/main.py
Yulia-Mala: ok

app/main.py
Yulia-Mala: NO. "car should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark". but whether condition meet  or not - could be realised it other method.

app/main.py
SergiiMachulin: There is link in my comment above, if you press PEP8

app/main.py
SergiiMachulin: Agree

SergiiMachulin (approved):

app/main.py
Serhii-Leonenko: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 in_class: int,
+                 in_mark: int,
+                 in_brand: str) -> None:
```
Serhii-Leonenko, line 5: Use the same names for the parameters and attributes. Also in other similar cases

Serhii-Leonenko (changes requested):

app/main.py
Yulia-Mala: additional check: done. regarding same names for the parameters and attributes - could you pls tell me why? I use "in_arg" names in funcs for clear understanding which data came from the outside

PR #383, opened 2023-02-05
Solution

SergiiMachulin:


kuznetsb (changes requested):
method, that washes a single car if wash_station.clean_power is greater than car.clean_mark
```
         if car.clean_mark <= self.clean_power:
```

kuznetsb (approved):
Nice work)

app/main.py
```diff
@@ -1,8 +1,82 @@
...
+
+        if not (isinstance(clean_mark, int)
+                and clean_mark in range(1, 11)):
+            raise ValueError(
+                "clean_mark argument must be from very dirty - 1 to clean - 10"
+            )
+        self.clean_mark = clean_mark
+
+        if not (isinstance(brand, str)):
+            raise TypeError("brand argument is not string")
```
Serhii-Leonenko, line 21: It's a redundant check for the brand because your type annotation says about that. Remove in all similar cases

app/main.py
```diff
@@ -1,8 +1,82 @@
...
+    def __init__(
+            self,
+            distance_from_city_center: int | float,
+            clean_power: int,
+            average_rating: int | float,
+            count_of_ratings: int
+    ) -> None:
+        if not (isinstance(distance_from_city_center, int | float)
+                and distance_from_city_center
+                in [value / 10.0 for value in range(1, 101, 1)]):
```
Serhii-Leonenko, line 35: Your function spends extra time creating a list and checking the value inside it. And it can be only float

app/main.py
```diff
@@ -1,8 +1,82 @@
...
+        if not (isinstance(clean_power, int)
+                and clean_power in range(1, 11)):
+            raise ValueError(
+                "clean_power argument must be from powerful 1 to powerless 10"
+            )
+        self.clean_power = clean_power
+
+        if not (isinstance(average_rating, int | float)
+                and average_rating
+                in [value / 10.0 for value in range(1, 51, 1)]):
```
Serhii-Leonenko, line 50: The same is here. And average rating could be only float

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #384, opened 2023-02-05
Solution

Litvinenko23:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return total_sum
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = (car.comfort_class * (self.clean_power - car.clean_mark)
+                * self.average_rating) / self.distance_from_city_center
+        cost = round(cost, 1)
+
+        return cost
+
+    def wash_single_car(self, car: Car) -> Car:
```
bihunva, line 38: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

bihunva (changes requested):

bihunva (approved):
Well done!

Serhii-Leonenko (approved):
Well done!

PR #388, opened 2023-02-05
Solution

ada-krav:


oksana0410 (approved):

app/main.py
```diff
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
Serhii-Leonenko, line 3: Indentation like this has better readability, fix it in all similar cases


```suggestion
    def __init__(self,
                comfort_class: int,
                clean_mark: int, 
                brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+            * self.average_rating
+            / self.distance_from_city_center), 1)
+
+    def rate_service(self, mark: int) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + mark)
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
+
+    def serve_cars(self, cars: list) -> float:
```
Serhii-Leonenko, line 30: You can give more descriptive type annotations, like this


```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        income = round(sum([
+            self.calculate_washing_price(car) for car in cars
+            if car.clean_mark < self.clean_power]), 1)
+        for car in cars:
+            self.wash_single_car(car)
```
Serhii-Leonenko, line 35: it's better to use one `for`  loop instead of both comprehension and for

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #389, opened 2023-02-05
Solution

Zenionoi:


app/main.py
```diff
@@ -1,8 +1,39 @@
...
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for i in cars:
+            if i.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(i)
+                self.wash_single_car(i)
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
```
BohdanLazaryshyn, line 32: you compare clean_mark and clean_power in the previous loop

BohdanLazaryshyn (changes requested):

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.average_rating = round(rating, 1)
+        self.count_of_ratings = count
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for i in cars:
```
yura-hudzovskyi, line 25: name "i" is a bad practice (checklist)
i -> car

app/main.py
```diff
@@ -1,8 +1,39 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(
+            self, distance: float, power: float, rating: float, count: int
+    ) -> None:
+        self.distance_from_city_center = distance
+        self.clean_power = power
+        self.average_rating = round(rating, 1)
```
amber-marichi, line 16: i am not sure it's necessary to use round() here since You do rounding in rate_service method as it should be

amber-marichi (changes requested):

amber-marichi (approved):
seems good. approved

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+            self, distance: float, power: float, rating: float, count: int
+    ) -> None:
+        self.distance_from_city_center = distance
+        self.clean_power = power
+        self.average_rating = rating
+        self.count_of_ratings = count
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
```
wQuelS, line 21: can you divide formula into smaller chunks?

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round((self.average_rating
```
wQuelS, line 36: same with dividing formula

wQuelS (changes requested):

app/main.py
Zenionoi: Not sure if I got it right, tried to following "good practice"

app/main.py
```diff
@@ -17,8 +17,11 @@ def __init__(
         self.count_of_ratings = count
 
     def calculate_washing_price(self, car: Car) -> float:
-        return round(car.comfort_class * (self.clean_power - car.clean_mark)
-                     * self.average_rating / self.distance_from_city_center, 1)
+        return round(car.comfort_class
+                     * (self.clean_power
```
wQuelS, line 21: Not it is formatted in right way, BUT.
In order to write more readable and maintainable code you should divide huge formula into small parts with using additional variables

app/main.py
```diff
@@ -34,6 +37,7 @@ def wash_single_car(self, car: Car) -> None:
 
     def rate_service(self, rate: int) -> None:
         self.average_rating = round((self.average_rating
-                                     * self.count_of_ratings + rate)
+                                     * self.count_of_ratings
```
wQuelS, line 40: same here. formatting is according to best practices, but it is better to decompose such huge formula

wQuelS (changes requested):

app/main.py
Zenionoi: Thanks, now I get it

YuriiKindrat (approved):

PR #390, opened 2023-02-06
Solution

BohdanLazaryshyn:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, num: int) -> float:
+        new_rating = round(((self.average_rating
+                             * self.count_of_ratings) + num)
+                           / (self.count_of_ratings + 1), 1)
+        self.average_rating = new_rating
+        self.count_of_ratings += 1
+        return self.average_rating
```
Danil228, line 48: I think you should not return anything in this method, because you are already changing the attribute

Danil228 (changes requested):

Danil228 (approved):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1
+                     )
```
YuriiKindrat, line 29: fix indentation here, or just split ur code to a several parts

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
```
wQuelS, line 26: can you just split your code into several smaller parts to make it more readable?

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, num: int) -> None:
+        new_rating = round(((self.average_rating
+                             * self.count_of_ratings) + num)
```
wQuelS, line 41: same, split your formula into smaller chunks, using additional variables

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+    def serve_cars(self, car_list: Car) -> float:
+        income = 0
+        for car in car_list:
+            if self.clean_power >= car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Serhii-Leonenko, line 38: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        difference = self.clean_power - car.clean_mark
+        upper_part = car.comfort_class * difference * self.average_rating
+        return round(upper_part / self.distance_from_city_center, 1)
+
+    def serve_cars(self, car_list: Car) -> float:
```
Serhii-Leonenko, line 29: It's incorrect type annotation

```suggestion
    def serve_cars(self, car_list: list[Car]) -> float:
```

Serhii-Leonenko (changes requested):

app/main.py
BohdanLazaryshyn: but we do a check before calling this method

Serhii-Leonenko (approved):
Well done!

PR #393, opened 2023-02-06
Solution car_wash_station

iiahnatiuk:


app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+        self, comfort_class: int, clean_mark: int, brand: str
```
oigod, line 3: Use code-style like in CarWashStation class (different code style)

oigod (changes requested):
Different code style, check comment under line 3

amber-marichi (changes requested):

amber-marichi (approved):
all seems good. approved

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        washing_price = car.comfort_class \
```
wQuelS, line 35: please, try to avoid line breaks. You can divide huge formula into smaller chunks via additional variables

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round(
+            ((self.average_rating * self.count_of_ratings) + rate)
+            / (self.count_of_ratings + 1), 1)
```
wQuelS, line 47: same

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        washing_price = car.comfort_class \
+            * (self.clean_power - car.clean_mark) \
+            * (self.average_rating / self.distance_from_city_center)
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
wQuelS, line 41: in your `serve_cars` method you already have a such check. So this method doesn't need it, since it will be called only if previous `if` statement is `True`

wQuelS (changes requested):

app/main.py
iiahnatiuk: done

app/main.py
iiahnatiuk: done

app/main.py
iiahnatiuk: done

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        sum_ratings = (self.average_rating * self.count_of_ratings) + rate
+        count_new_rating = self.count_of_ratings + 1
+        self.average_rating = round((sum_ratings / count_new_rating), 1)
+        self.count_of_ratings += 1
```
YuriiKindrat, line 47: why do u need count_new_rationg variable if u are making self.count_of_ratings += 1 ?

YuriiKindrat (changes requested):

app/main.py
iiahnatiuk: fixed

YuriiKindrat (approved):

PR #401, opened 2023-02-06
Solution: wash station

ZhAlexR:


app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
```
AnnaBoichenko, line 3: it`s unnecessary to leave pass

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
```
AnnaBoichenko, line 8: it`s unnecessary to leave pass

app/main.py
AnnaBoichenko: you can also delete a comment

app/main.py
AnnaBoichenko: you can also delete a comment

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+            if car.clean_mark < self.clean_power:
+                total_price += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return total_price
+
+    def rate_service(self, rate_mark: float) -> None:
+        self.average_rating = round((
+            (self.count_of_ratings * self.average_rating + rate_mark)
+            / (self.count_of_ratings + 1)), 1)
+        self.count_of_ratings += 1
```
AnnaBoichenko, line 46: no newline at end of the file

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+            car.clean_mark = self.clean_power
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if self.clean_power > car.clean_mark:
+            return round((car.comfort_class
+                          * (self.clean_power - car.clean_mark)
+                          * self.average_rating
+                          / self.distance_from_city_center), 1)
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
AnnaBoichenko, line 34: TypeError: 'type' object is not subscriptable
please, instead try cars: list

AnnaBoichenko (changes requested):

app/main.py
ZhAlexR: @AnnaBoichenko
Sorry, but this red  area means that the rows were by default in the code.
Please, take a look to the green area.

app/main.py
ZhAlexR: @AnnaBoichenko The same. It is default version. The red area does not exist in my code version.

app/main.py
ZhAlexR: @AnnaBoichenko do not clearly understand your point.
I used this [example](https://docs.python.org/3/library/typing.html#:~:text=type%20inside%20TypeGuard.-,for%20example%3A,-def%20is_str_list(val) to format my code. 
Could you clarify your POV, please?

ZhAlexR: @AnnaBoichenko Do not understand your comment.
Could you clarify your POV, please?
I used [this example](https://docs.python.org/3/library/typing.html#:~:text=type%20inside%20TypeGuard.-,for%20example%3A,-def%20is_str_list(val) to format my code

app/main.py
YuriiKindrat, line 46: keep ur code DRY, u can make self.count_of_ratings += 1 only once

YuriiKindrat (changes requested):

AnnaBoichenko (approved):

YuriiKindrat (approved):

PR #404, opened 2023-02-06
Solution

bohdan-sain:


ZhAlexR (changes requested):

ZhAlexR (approved):

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = round(car.comfort_class
+                      * (self.clean_power - car.clean_mark)
+                      * self.average_rating
+                      / self.distance_from_city_center, 1)
+        return price
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
wQuelS, line 38: in `serve_cars` you already have the same check. And the car will be washed only if the statement is `True`

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
wQuelS, line 22: you can use more secriptive type hints, for example:
cars: List[Car]

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = round(car.comfort_class
+                      * (self.clean_power - car.clean_mark)
```
wQuelS, line 32: Can you divide such huge formula into smaller chunks?

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                      / self.distance_from_city_center, 1)
+        return price
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> (float, int):
+        self.average_rating = round((self.count_of_ratings
+                                    * self.average_rating
```
wQuelS, line 43: same with dividing huge formula

YuriiKindrat (approved):

PR #406, opened 2023-02-06
Solution

ykopatko:


VladyslavCherkashyn (approved):
Well done!

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * (self.average_rating / self.distance_from_city_center), 1)
```
YuriiKindrat, line 25: fix identation here, or just split ur code to a several parts

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
YuriiKindrat, line 35: the same here, also keep ur code DRY - no need to make (self.count_of_ratings + 1)  if u make self.count_of_ratings += 1 later

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for i in cars:
```
YuriiKindrat, line 39: i is a bad name for variable, make the name for variable more descriptive

YuriiKindrat (changes requested):

app/main.py
ykopatko: if I use  self.count_of_ratings += 1 before recalculating of the new average rating, I will need to use (self.average_rating * (self.count_of_ratings - 1)+ rate), which seems less comfortable. Becase in such a case count of ratings would be higher on 1 point. Does it make sense?

wQuelS (approved):
GJ

PR #410, opened 2023-02-06
Solution

tytomyr:


app/main.py
```diff
@@ -21,15 +23,19 @@ def serve_cars(self, cars: list[Car]) -> float:
             self.wash_single_car(i)
         return income
 
-    def calculate_washing_price(self, car: Car):
-        return round(car.comfort_class * (self.clean_power - car.clean_mark) * (
-                    self.average_rating / self.distance_from_city_center), 1)
+    def calculate_washing_price(self, car: Car) -> float:
+        one = self.clean_power - car.clean_mark
```
ZhAlexR, line 27: It would be nice to use more meaningful variables name. For example, "clean_level_difference" instead of "one", or "rating_dist _ratio" instead of "two"

app/main.py
```diff
@@ -21,15 +23,19 @@ def serve_cars(self, cars: list[Car]) -> float:
...
-    def wash_single_car(self, car: Car):
+    def wash_single_car(self, car: Car) -> None:
         if self.clean_power >= car.clean_mark:
             car.clean_mark = self.clean_power
 
     def rate_service(self, rate: int) -> None:
-        self.average_rating = round((self.average_rating * self.count_of_ratings + rate) / (self.count_of_ratings + 1),
-                                    1)
+        three = self.average_rating * self.count_of_ratings + rate
+        four = self.count_of_ratings + 1
```
ZhAlexR, line 38: More meaningful variables name

app/main.py
```diff
@@ -21,15 +23,19 @@ def serve_cars(self, cars: list[Car]) -> float:
             self.wash_single_car(i)
         return income
```
ZhAlexR, line 25: It would be better to use "car" instead of "i" in for-loop

ZhAlexR (changes requested):

ZhAlexR (approved):

app/main.py
YuriiKindrat: i is a bad name for variable, use more descriptive names for variables

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+            if i.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(i)
+            self.wash_single_car(i)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_power_difference = self.clean_power - car.clean_mark
+        rating_ration = self.average_rating / self.distance_from_city_center
+        return round(
+            car.comfort_class * clean_power_difference * rating_ration, 1)
```
YuriiKindrat, line 30: fix indentation here

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        count_ration = self.average_rating * self.count_of_ratings + rate
+        rating_ration = self.count_of_ratings + 1
+        self.average_rating = round(
+            count_ration / rating_ration, 1)
```
YuriiKindrat, line 40: the same here

YuriiKindrat: also keep ur code DRY no need to make this - (self.count_of_ratings + 1) if u make this before self.count_of_ratings += 1

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center, 1
```
wQuelS, line 29: split formula into smaller chunks to increase readability

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
YuriiKindrat, line 3: it is better to use one line per one argument for better readability

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
YuriiKindrat, line 12: the same here

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Well done!

PR #411, opened 2023-02-06
Solution

ProdyRodion:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car: list) -> float:
```
ramsty, line 22: Here we have a list of cars, so I'd be better use "cars" for name of the list and "car" when iterate through list.

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                self.wash_single_car(one)
+        return price
+
+    def calculate_washing_price(self, car: Car) -> float:
+        wash_price = round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * (self.average_rating / self.distance_from_city_center), 1)
+        return wash_price
+
+    def wash_single_car(self, car: Car) -> float:
```
ramsty, line 36: This func doesn't return anything -> None
Also "pass" isn't necessary

PR #418, opened 2023-02-06
'Solution'

oksanaaam:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0.0
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return round(income, 1)  # pytest
```
Bohdan-Kovalenko1992, line 33: Remove unwanted comments

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_car = self.clean_power - car.clean_mark
+        service_level = clean_car * car.comfort_class  # service_level car
```
Bohdan-Kovalenko1992, line 37: same

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        quality_service = service_level * self.average_rating
+        price = quality_service / self.distance_from_city_center
+        return price
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
```
Bohdan-Kovalenko1992, line 43: it will be true if  `self.clean_power == car.clean_mark`
I don't think this is how it should work

Bohdan-Kovalenko1992 (approved):
Good job!

app/main.py
YuriiKindrat: remove unnecessary comments

app/main.py
YuriiKindrat: +

app/main.py
YuriiKindrat: +

YuriiKindrat (changes requested):

app/main.py
oksanaaam: It's working like before, because without if `self.clean_power == car.clean_mark`  instead of `self.clean_power >= car.clean_mark` the program has 3 failed tests, you can check my code in your IDE.
About removing unnecessary comments - OK, I'll remove it.

Bohdan-Kovalenko1992: i mean its u have to use > or <, not >= or <=. Because in situation when `self.clean_power == car.clean_mark` ur program will work wrong.

wQuelS (approved):
Looks nice

PR #421, opened 2023-02-07
Solution

dogmoony:


ProdyRodion (approved):
good job

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        clean_difference = self.clean_power - car.clean_mark
+        comfort_rating = car.comfort_class * self.average_rating
+        comfort_dist = comfort_rating / self.distance_from_city_center
+        price = comfort_dist * clean_difference
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark <= self.clean_power:
+            car.clean_mark = self.clean_power
+            return True
```
wQuelS, line 39: Do you really need this `return True`?

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_difference = self.clean_power - car.clean_mark
+        comfort_rating = car.comfort_class * self.average_rating
+        comfort_dist = comfort_rating / self.distance_from_city_center
+        price = comfort_dist * clean_difference
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark <= self.clean_power:
```
wQuelS, line 37: you have a pretty similar check in your `serve_cars` method. Maybe you can somehow avoid this code duplication?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) -> any:
```
YuriiKindrat, line 2: Are u sure, that init method returns any ?

YuriiKindrat (changes requested):

Serhii-Leonenko (changes requested):
All checks should be passed

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_difference = self.clean_power - car.clean_mark
+        comfort_rating = car.comfort_class * self.average_rating
+        comfort_dist = comfort_rating / self.distance_from_city_center
+        price = comfort_dist * clean_difference
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Serhii-Leonenko, line 39: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Serhii-Leonenko, line 23: You can use more descriptive type annotations


```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+        self, comfort_class: int, clean_mark: int, brand: str
+    ) -> None:
```
Serhii-Leonenko, line 4: It's ok, but it will be better if you use the same style for parameters according to the init in CarWashStation, it has better readability.

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #425, opened 2023-02-07
Solution

myroslavatrschk:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income_for_wash += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income_for_wash
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round((car.comfort_class
+                     * (self.clean_power
+                        - car.clean_mark)
```
wQuelS, line 37: Can you split such huge formula into smaller chunks to increase readability of your code?

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                     * self.average_rating
+                     / self.distance_from_city_center), 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
wQuelS, line 42: Don't you have same check in your `serve_cars` method?

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(((self.average_rating
+                                      * self.count_of_ratings
+                                      + rate)
```
wQuelS, line 48: same about readability

wQuelS (changes requested):

myroslavatrschk (commented):
I tried to break huge formulas by adding new variables. I'm not sure if it's correct, but this was my first thought.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> any:
```
YuriiKindrat, line 20: are u sure about that init method returns any ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #427, opened 2023-02-07
'Solution'

RuslanLopatin:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                      for car in washed_cars])
+        for car in washed_cars:
+            self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: str) -> None:
+        cost = car.comfort_class \
+               * (self.clean_power - car.clean_mark)\
+               * self.average_rating \
+               / self.distance_from_city_center
```
amber-marichi, line 33: maybe you should separate this into few more concise formulas to improve code readability. Also maybe you should check the task, it seems like this method can be called outside of serve_cars so it also need to check clean_power/clean_mark before calculations

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+    def wash_single_car(self, car: str) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
+                                     * (self.count_of_ratings - 1) + rating)
+                                    / self.count_of_ratings, 1)
```
amber-marichi, line 44: same here. maybe it is good idea to write result of this formula in another variable so the formula could be separated from round() function. please check

amber-marichi (changes requested):

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                       < self.clean_power]
+        income = sum([self.calculate_washing_price(car)
+                      for car in washed_cars])
+        for car in washed_cars:
+            self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: str) -> None:
+        cost = car.comfort_class \
+               * (self.clean_power - car.clean_mark)\
```
destr34, line 31: you should add variables to your formula for better readability 
like subtraction = a-b

Serhii-Leonenko (changes requested):
All checks should be passed

app/main.py
```diff
@@ -1,8 +1,39 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: str,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Serhii-Leonenko, line 4:
```suggestion
    def __init__(self, 
                 comfort_class: str,
                 clean_mark: int,
                 brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
Serhii-Leonenko, line 13: Fix indentation according to previous comment

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Serhii-Leonenko, line 19: You can give more descriptive type annotations

```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        washed_cars = [car for car in cars
+                       if car.clean_mark < self.clean_power]
+        income = sum(self.calculate_washing_price(car) for car in washed_cars)
+        for car in washed_cars:
+            self.wash_single_car(car)
+        return round(income, 1)
```
Serhii-Leonenko, line 25: it's redundant to use both the `for` loop and comprehension here. Your code spends extra time on it.

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        return round(part1 * part2, 1)
+
+    def wash_single_car(self, car: str) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> None:
+        self.count_of_ratings += 1
+        obj1 = self.average_rating * (self.count_of_ratings - 1) + rating
+        obj2 = self.count_of_ratings
+        self.average_rating = round(obj1 / obj2, 1)
```
Serhii-Leonenko, line 39: Try to use more descriptive names for your variables.
Also, obj2 is redundant variable

Serhii-Leonenko (changes requested):

amber-marichi (approved):

app/main.py
YuriiKindrat: fix this

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+    def serve_cars(self, cars: list) -> float:
+        washed_cars = [car for car in cars
+                       if car.clean_mark < self.clean_power]
+        income = sum(self.calculate_washing_price(car) for car in washed_cars)
+        [self.wash_single_car(car) for car in washed_cars]
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: str) -> float:
+        part1 = car.comfort_class * (self.clean_power - car.clean_mark)
+        part2 = self.average_rating / self.distance_from_city_center
```
YuriiKindrat, line 31: use more descriptive names for variables - they should describe what are u calculating

YuriiKindrat (changes requested):

amber-marichi (approved):

app/main.py
wQuelS: You still didn't fix this

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        washed_cars = [car for car in cars
+                       if car.clean_mark < self.clean_power]
+        income = sum(self.calculate_washing_price(car) for car in washed_cars)
+        [self.wash_single_car(car) for car in washed_cars]
```
Serhii-Leonenko, line 26: You don't need three comprehensions here. Remove redundant or make the same with only one for loop

app/main.py
Nattalli: look: type list can have a lot of different types inside. For example, a list of integers, strings, and Car class instances. So we recommend clarifying the list type: `list[int]`, `list[Car]`, ...
In this case, it is the list of Car, so: `list[Car]`

Nattalli (approved):

PR #444, opened 2023-02-21
'Solution'

kirzenko:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
vasylhnatiuk, line 34: you must add condition that self.clean_power > car.clean_mark

app/main.py
```diff
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
vasylhnatiuk, line 2: would be better to use  
```
 def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) 
```

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
```
vasylhnatiuk, line 12: same here

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
```
vasylhnatiuk, line 20: list of what? please add a more accurate annotation

vasylhnatiuk (changes requested):

PR #450, opened 2023-02-24
Done

MaksNochvai:


Bohdan-Salamakha (changes requested):
don't work on the master branch - bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #457, opened 2023-03-03
draft

sbilikepy:
skip it please :)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,8 @@
 class Car:
     def __init__(self,
-                 comfort_class: int = 7,
-                 clean_mark: int = 1,
-                 brand: str = "Unknown"
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
                  ) -> None:
         self.comfort_class = comfort_class
         self.clean_mark = clean_mark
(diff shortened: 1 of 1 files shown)


sbilikepy:
draft PR comment test
 ／l、
（ﾟ､ ｡７
⠀ l、ﾞ~ヽ 
    じし(_,  )ノ

sbilikepy:
draft PR comment test
/ᐠ. .ᐟ\ฅ

ArsenPidhoretskyi (changes requested):
Draft

vasylhnatiuk (changes requested):
flake8 tests failed

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        cars_to_wash = [car for car in cars
+                        if car.clean_mark < self.clean_power]
+        for car in cars_to_wash:
+            income += self.calculate_washing_price(car)
```
vasylhnatiuk, line 29: write `if car.clean_mark < self.clean_power` here, and delete 26-27 lines. it is better to use minimum `for` loops

app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int = 7,
```
vasylhnatiuk, line 3: why do you add default value for each argument? delete this here and everywhere. it must be only `comfort_class: int`

vasylhnatiuk (changes requested):

app/main.py
sbilikepy: To maximize profit if we get a broken class instance :)

app/main.py
sbilikepy: Got it

ArsenPidhoretskyi (approved):
Well done!

PR #463, opened 2023-03-05
Develop

PetroHavrysh:


app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                self.wash_single_car(car)
+        return round(total_income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.dis_to_city, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
ArsenPidhoretskyi, line 31: > wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark

Checks does not exist

ArsenPidhoretskyi (changes requested):

app/main.py
PetroHavrysh: а хіба тут це не перевіряється: if car.clean_mark < self.clean_power: у def serve_cars(self, cars: list) -> float:

Grayder0152 (approved):

PR #482, opened 2023-03-06
Task

n1ckmel:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center),
+                1)
+        return 0
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars_list: list) -> float:
```
ArsenPidhoretskyi, line 44: list of what?
Add inner type

ArsenPidhoretskyi: Also rename to cars no need _list suffix

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
```
ArsenPidhoretskyi, line 33: Nothing said about condition

> calculate_washing_price - method, that calculates cost for a single car wash, cost is calculated as: car's comfort class * difference between wash station's clean power and car's clean mark * car wash station rating / car wash station distance to the center of the city, returns number rounded to 1 decimal;

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #490, opened 2023-03-15
'WashingStation'

NataliaStepanenko:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            clean_power: int,
+            average_rating: int,
+            count_of_ratings: int,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list) -> float:
```
ArsenPidhoretskyi, line 26: list_of_cars
Why you need prefix list_????
rename to cars

ArsenPidhoretskyi: Also list of what provide inner type

ArsenPidhoretskyi (changes requested):

app/main.py
NataliaStepanenko: ok, thanks.

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #496, opened 2023-03-22
Solution py-car-wash-station

Fawnygod:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark > self.clean_power:
+                break
```
ArsenPidhoretskyi, line 29: Why we need to do break in this case? We do not need to calculate only and wash

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return ((car.comfort_class * (self.clean_power - car.clean_mark)
+                 ) * self.average_rating) / self.distance_from_city_center
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
ArsenPidhoretskyi, line 40: If-statement does not exist 

> wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark

ArsenPidhoretskyi (approved):
Well done!

PR #501, opened 2023-03-28
Solution

alim2709:
пайчарм подсвечивает некоторые моменты, но пайтест проходит 
image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(
+            self, distance_from_city_center: (int or float), clean_power: int,
+            average_rating: float, count_of_ratings: int
+    ) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        cost = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                cost += self.calculate_washing_price(car)
+                car.clean_mark = self.wash_single_car(car)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(
+            self, distance_from_city_center: (int or float), clean_power: int,
```
Bohdan-Salamakha, line 14: distance_from_city_center: (int or float) is a bad typing
use:
distance_from_city_center: int | float      for python 3.10+
or
distance_from_city_center: Union[int, float]

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            average_rating: float, count_of_ratings: int
+    ) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Bohdan-Salamakha, line 23: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        cost = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                cost += self.calculate_washing_price(car)
+                car.clean_mark = self.wash_single_car(car)
+        return round(cost, 1)
+
+    def calculate_washing_price(self, car: object) -> float:
```
Bohdan-Salamakha, line 31: car: object is a non-informical annotation cause every object in Python is inherited from object
use car: Car annotation maybe

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+        return (car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center)
+
+    def wash_single_car(self, car: object) -> object:
```
Bohdan-Salamakha, line 38: car: object is a non-informical annotation cause every object in Python is inherited from object
use car: Car annotation maybe

rewrite return typehint also

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #509, opened 2023-04-08
Solution of task "Car and WashStation."

laskarj:
I didn't understand where exactly the serve_cars function should be, so I did both, but the task says it's a method, but it was called as a function in the example, which is a bit confusing.
And in the tests it is called both as a method and as a function, so I just did both.
But I made serve_cars as a function a little differently so as not to rewrite it every time for a new car wash.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,73 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
+
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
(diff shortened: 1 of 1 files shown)


Vasyl-Smutok:
image

app/main.py
```diff
@@ -1,8 +1,73 @@
...
+    but the tests fail if it is only in the middle of the class.
+    I've deviated a bit from the task, because it seems to me that
+    if this function should be outside the class, we should specify
+    and station in the arguments to avoid rewriting the code every time.
+    """
+    for car in cars:
+        if car.clean_mark < wash_station.clean_power:
+            income += wash_station.calculate_washing_price(car)
+            wash_station.wash_single_car(car)
+    return round(income, 1)
```
Vasyl-Smutok, line 73: Remove it

Vasyl-Smutok (changes requested):

app/main.py
laskarj: Done.

Vasyl-Smutok (approved):
Good Job 👍

PR #512, opened 2023-04-15
Solution v1

neostyle88:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Bohdan-Salamakha, line 23: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        money = 0
+        for car in cars:
+            money += self.wash_single_car(car)
+        return money
```
Bohdan-Salamakha, line 27: use list comprehension to do it

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        money = 0
+        income = (money + self.wash_single_car(car) for car in cars)
+        return sum(income)
```
Bohdan-Salamakha, line 26:
```suggestion
        return sum(self.wash_single_car(car) for car in cars)
```

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #514, opened 2023-04-17
Solution

KittNeKit:


app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str
```
EdAlekseiev, line 4: Add trailing comma, please

app/main.py
```diff
@@ -1,8 +1,46 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
```
EdAlekseiev, line 17: Add trailing comma

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
EdAlekseiev, line 24: Please, improve type hint for `cars`. Use `list[Car]` instead of `list`

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        return sum(self.wash_single_car(car) for car in cars)
```
EdAlekseiev, line 25: We can wash the car only if `clean_mark < clean_power`. You must filter cars in this method to avoid unnecessary calls of `wash_single_car`.

EdAlekseiev (changes requested):

app/main.py
KittNeKit: In `wash_single_car` I filter cars, so I can use the `wash_single_car` method to simplify and shorten the code, I should rewrite the code, or let it be?

EdAlekseiev: Yes, rewrite it, please

EdAlekseiev (changes requested):

app/main.py
KittNeKit: I had an error "AssertionError: Expected '`wash_single_car`' to have been called inside '`serve_cars`' method", so even pytest says, that we need to use `wash_single_car` in `serve_cars`.

Bohdan-Salamakha (approved):

PR #532, opened 2023-05-05
Car wash station functionality implemented

thorium-zero:


Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        for car in cars:
+            income += self.wash_single_car(car)
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        if self.clean_power <= car.clean_mark:
+            return 0
+        income = self.calculate_washing_price(car)
+        car.clean_mark = self.clean_power
+        return income
```
Oleg-s-s, line 33: No need to calculate income here, just check if condition is True and set the clean mark. Also according to task: 
_if wash_station.clean_power is greater than car.clean_mark_ is a strict requirement

Oleg-s-s (changes requested):

Oleg-s-s (approved):

PR #540, opened 2023-05-11
Create washing station

Bohdan11Dii:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car: list) -> float:
```
Bohdan-Salamakha, line 22: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

PR #549, opened 2023-05-16
Solution

prykhodkom:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
YukihiroSM, line 44: You are using round with separate variable in `calculate_washing_price`, but doing everything as "oneliner"  in rate service.
Code style should be same.

PR #551, opened 2023-05-19
implemented automated calculation of prices for car wash

Oknametra:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        total_cost = 0
+        for car in cars:
+            total_cost += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return round(total_cost, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if self.clean_power > car.clean_mark:
+            return car.comfort_class * (self.clean_power - car.clean_mark)\
+                * self.average_rating / self.distance_from_city_center
```
ArsenPidhoretskyi, line 29:
```suggestion
            return (
                car.comfort_class * (self.clean_power - car.clean_mark)
                * self.average_rating / self.distance_from_city_center
            )
```

ArsenPidhoretskyi: Bette to use brackets instead of \

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        else:
```
ArsenPidhoretskyi, line 30: else: is not needed in this case

app/main.py
Oknametra: kk, thx

app/main.py
Oknametra: got it

PR #562, opened 2023-05-26
Solution

YKonoplyov:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * (self.average_rating / self.distance_from_city_center),
+            1
+        )
+        return price
```
Oleg-s-s, line 28: If u used a round() function instantly there is no need in using additional variable then

Oleg-s-s (changes requested):

YuriiKindrat (approved):

PR #569, opened 2023-05-28
Solution by Alex

AlexIzmashkin:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
Bohdan-Salamakha, line 12:
```suggestion
    def __init__(self,
                 distance_from_city_center: float,
                 clean_power: int,
                 average_rating: float,
                 count_of_ratings: int
                 ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Bohdan-Salamakha, line 18: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

Bohdan-Salamakha (changes requested):

PR #573, opened 2023-05-29
Solution

mwellick:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
Bohdan-Salamakha, line 14:
```suggestion
    def __init__(
                 self,
                 distance_from_city_center: float,
                 clean_power: int,
                 average_rating: float,
                 count_of_ratings: int
                 ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        income = round(car.comfort_class * (self.clean_power - car.clean_mark)
+                       * self.average_rating
+                       / self.distance_from_city_center, 1)
+        return income
```
Bohdan-Salamakha, line 32: u don't need to create this variable here
just return rounded value

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #575, opened 2023-05-29
Solution

Po-SvinSkipodla:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                res += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return round(res, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+
+        return car.comfort_class * (self.clean_power - car.clean_mark) \
```
LiudmylaKulinchenko, line 39: Please, don't use `* \` , you can add W504 flake8 rule to ignored ones or create a few temporal variables

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1),
+            1)
```
LiudmylaKulinchenko, line 50: The braces should be on the same level and start with the text or have a line break between the text

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 😎

PR #581, opened 2023-05-29
Solution

AntonShpakovych:
reopen pull request for updating students-table


Oleg-s-s (approved):
Interesting solution for validation, but __ROUNDING_PRECISION I think is redundant here

PR #587, opened 2023-05-29
Solution

Kasatello:


app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
YuriiKindrat, line 4: it is better to place one arg per line for better readability )

YuriiKindrat (approved):

app/main.py
Kasatello: кк

PR #588, opened 2023-05-29
"Solution"

HDKESL:


app/main.py
```diff
@@ -1,8 +1,47 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: str,
+                 clean_mark: float,
+                 brand: str) -> None:
```
YuriiKindrat, line 4: it is better to write one arg per line for better readability )

YuriiKindrat (approved):

PR #589, opened 2023-05-29
my_solution

Mr-Furman:


app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
YuriiKindrat, line 4: it is better to write only one arg per line for better readability, fix this everywhere)

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+             + rate)
+            / self.count_of_ratings), 1)
+        return self.average_rating
+
+    def serve_cars(self, cars_list: list) -> float:
+        serve_for_cars = [self.calculate_washing_price(car)
+                          for car in cars_list
+                          if car.clean_mark <= self.clean_power]
+        for car in cars_list:
+            self.wash_single_car(car)
```
YuriiKindrat, line 43: why do u need 2 loops for calculating this ? maybe it is better to use for loop for both cases ?

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+    def rate_service(self, rate: int) -> float:
+        self.count_of_ratings += 1
+        self.average_rating = round((
+            ((self.count_of_ratings - 1)
+             * self.average_rating
+             + rate)
+            / self.count_of_ratings), 1)
+        return self.average_rating
+
+    def serve_cars(self, cars_list: list) -> float:
```
YuriiKindrat, line 38: it is better to use more descriptive type annotation f.e. list[Car]

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job

PR #619, opened 2023-06-20
Solution

pnakongit:


app/main.py
```diff
@@ -1,8 +1,69 @@
...
+        div_rating = self.average_rating / self.distance_from_city_center
+        income = car.comfort_class * difference_power_mark * div_rating
+
+        return round(income, 1)
+
+    def can_wash_car(self, car: Car) -> bool:
+        if self.clean_power > car.clean_mark:
+            return True
+
+        return False
```
ArsenPidhoretskyi, line 54:
```suggestion
        return self.clean_power > car.clean_mark:
```
You can do this inline

ArsenPidhoretskyi (changes requested):

app/main.py
pnakongit: Noted! Thx a lot

ArsenPidhoretskyi (approved):
Well done!

PR #624, opened 2023-06-22
'Solution'

Rudge0:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
vasylhnatiuk, line 29:
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        total_price = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total_price += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return total_price
+
+    def calculate_washing_price(self, car: Any) -> float:
```
vasylhnatiuk, line 37:
```suggestion
    def calculate_washing_price(self, car: Car) -> float:
```

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        price = car.comfort_class * \
+            ((self.clean_power - car.clean_mark)
+                * self.average_rating / self.distance_from_city_center)
+
+        return round(price, 1)
+
+    def wash_single_car(self, car: str) -> None:
```
vasylhnatiuk, line 44: also wrong annotation

vasylhnatiuk (changes requested):

app/main.py
Rudge0: what`s wrong here?

ArsenPidhoretskyi (approved):
Well done!

PR #635, opened 2023-06-25
Solution

fotomassimo:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
Nattalli, line 11: not an error, but a recommendation. It is more readable to write in this way:
```python
def __init__(
    self, 
    clean_power: int,
    average_rating: float, 
    count_of_ratings: int
) -> None:

Comment (Nattalli) [reply app/main.py:11]: the same for every function and method where you have a lot of args and you can't take it in 1 line

File: app/main.py [hunk 2]
```diff
@@ -1,8 +1,41 @@
...
+    def calculate_washing_price(self, car_item: Car) -> float:
+        washing_price = car_item.comfort_class \
+            * (self.clean_power - car_item.clean_mark) \
+            * self.average_rating \
+            / self.distance_from_city_center
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car_item: Car) -> None:
+        car_item.clean_mark = self.clean_power
+        pass
```
Nattalli, line 26: please, remove `pass` statement from every method, it is unused (we add it to show you that you should implement these methods)

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+    def rate_service(self, rating: int) -> None:
+        new_average_rating = (self.average_rating * self.count_of_ratings
+                              + rating) / (self.count_of_ratings + 1)
+        self.average_rating = round(new_average_rating, 1)
+        self.count_of_ratings += 1
+        pass
+
+    def serve_cars(self, cars_list: list) -> float:
+        total_sum = 0
+        for car_item in cars_list:
```
Nattalli, line 37:
```python
for car in cars_list
```
seems a little better

Nattalli (changes requested):

Nattalli (approved):

PR #640, opened 2023-06-28
Solution-Milokumov

EvgenMilokumov:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = car.comfort_class * \
+            ((self.clean_power - car.clean_mark) * self.average_rating
+             / self.distance_from_city_center)
```
vsmutok, line 33: better to use `()` then `\`
fix everywhere

vsmutok (changes requested):

app/main.py
EvgenMilokumov: done

vsmutok (approved):
Looks Good 👍

PR #642, opened 2023-07-03
Implemented py-car-wash-station

katryana:
Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,9 +2,9 @@ class Car:
     def __init__(self, comfort_class: int,
                  clean_mark: int, brand: str) -> None:
         if comfort_class < 1 or comfort_class > 7:
-            return TypeError
+            raise TypeError
         if clean_mark < 1 or clean_mark > 10:
-            return TypeError
+            raise TypeError
         self.comfort_class = comfort_class
         self.clean_mark = clean_mark
         self.brand = brand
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,36 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) -> None:
```
Kroll410, line 2: We need to have validation for these values based on task description

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
```
Kroll410, line 9: Validation

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car_cal: Car):
+        price = car_cal.comfort_class * (self.clean_power - car_cal.clean_mark)
+        price *= self.average_rating / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car_wash: Car):
```
Kroll410, line 21: rename the variables for better understanding its purpose. Here it can be simply `dirty_car`

Kroll410, line 21: rename the variables for better understanding its purpose. Here it can be simply `dirty_car`

Kroll410 (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
Kroll410, line 2: You should not print the exceptions, but raise them

Kroll410 (changes requested):

Kroll410 (changes requested):

Kroll410 (approved):

PR #661, opened 2023-07-17
Solution

WoyDiam:


tests/test_main.py
```diff
@@ -35,7 +35,7 @@ def test_car():
         ),
     ],
 )
-def test_car_wash_station(cars, wash_station, total_cost):
+def test_car_wash_station(cars: object, wash_station: object, total_cost: object) -> object:
```
ArsenPidhoretskyi, line 38: No need to change this file

ArsenPidhoretskyi (changes requested):
Fix flake8

PR #671, opened 2023-07-25
Created class Car and implemented class CarWashStation with 4 methods as per task

sberdianskyi:


ArsenPidhoretskyi (approved):
Well done!

ArsenPidhoretskyi (approved):

PR #677, opened 2023-08-01
Classes for car station

maxkatkalov:
Solution implemented, please check.

And answer: is it a good idea to add doc-strings to the methods in this case?


app/main.py
```diff
@@ -1,8 +1,80 @@
...
+
+    def wash_single_car(self, car: str) -> None:
+        """
+        method, that washes a single car,
+        so it should have clean_mark equals
+        wash station's clean_power, if
+        wash_station.clean_power is greater
+        than car.clean_mark;
+        """
+        if car.clean_mark < self.clean_power:
```
nyanplague, line 66: maybe this line is unnecessary because of `if` on line 53?

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+
+        result = 0
+
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                result += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(result, 1)
+
+    def wash_single_car(self, car: str) -> None:
```
nyanplague, line 58: car annotation is not str, car is type of Car :)

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        """
+
+        return round(
+            car.comfort_class
+            * abs(self.clean_power - car.clean_mark)
+            * (self.average_rating / self.distance_from_city_center),
+            1,
+        )
+
+    def serve_cars(self, cars: list[str]) -> float:
```
nyanplague, line 42: same as on line 58

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        count_of_ratings: int,
+        sum_of_ratings: int = 0,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+        self.sum_of_ratings = average_rating * count_of_ratings
+
+    def calculate_washing_price(self, car: str) -> float:
```
nyanplague, line 25: same as on line 58

app/main.py
```diff
@@ -1,8 +1,80 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int,
+        sum_of_ratings: int = 0,
```
nyanplague, line 17: This attribute is unused, because sum of ratings is calculated on line 23

nyanplague (changes requested):
Left few small suggestions, otherwise good job 👍

app/main.py
maxkatkalov: ok, agree

app/main.py
maxkatkalov: Sure, I edited it everywhere. Thank u.

Kroll410 (approved):

PR #682, opened 2023-08-02
Solution

nyanplague:


KristianYe (approved):
Cool solution with exceptions) 👍

app/main.py
```diff
@@ -1,8 +1,84 @@
+class CarValidationError(Exception):
+    """This Exception is for errors when initializing Car object"""
+
+
+class CarWashStationError(Exception):
+    """This Exception is for errors when initializing CarWashStation object"""
```
Kroll410, line 6: In practice, if your exceptions are not self-explanatory, you don't need to pass a doc. But here I guess we need to, but write a doc from a 3d person perspective (there's no `This`)
There's error with initialization of  ...

Kroll410 (changes requested):

app/main.py
nyanplague: Thank you!

PR #687, opened 2023-08-04
'Solution'

lgSilay:


eduardhabryd (changes requested):
There is a better way to add docstrings: https://www.jetbrains.com/help/pycharm/creating-documentation-comments.html

Here is an example:

        """
        Method, that calculates cost for a single car wash

        :param car: a car that we need to calculate the price for
        :return: returns number rounded to 1 decimal
        """

eduardhabryd (approved):

Kroll410 (approved):

PR #688, opened 2023-08-04
Solution

eduardhabryd:


app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        for serving this list of Car's,
+        rounded to 1 decimal
+        """
+        income = 0
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return round(income, 1)
```
maxkatkalov, line 38: It can be implemented in one line return, it is more efficient. Try to make calculations in `return round()`

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        """
+        Method, that washes a single car
+
+        :param car: a car that we want to wash
+        :return: it should have clean_mark equals wash station's clean_power,
+        if wash_station.clean_power is greater than car.clean_mark
+        """
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+        return None
```
maxkatkalov, line 64: Unnecessary return here, because annotation `-> None` doesn't make u to write return

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        :return: based on this single rate average_rating
+        and count_of_ratings should be changed
+        """
+        new_average = round(
+            (self.count_of_ratings * self.average_rating + rating)
+            / (self.count_of_ratings + 1),
+            1)
+        self.count_of_ratings += 1
+        self.average_rating = new_average
+        return None
```
maxkatkalov, line 80: Unnecessary return here, because annotation `-> None` doesn't make u to write return

maxkatkalov (changes requested):
Good job, but some small changes are needed to be implemented to clean code.

app/main.py
```diff
@@ -1,8 +1,78 @@
...
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        """
+        Method, that calculates cost for a single car wash
+
+        :param car: a car that we need to calculate the price for
+        :return: returns number rounded to 1 decimal
+        """
+        power_mark_difference = self.clean_power - car.clean_mark
```
maxkatkalov, line 47: It can be implemented in one line return, it is more efficient. Try to make calculations in `return round()`

app/main.py
```diff
@@ -1,8 +1,78 @@
...
+        washing_price = (car.comfort_class
+                         * power_mark_difference
+                         * self.average_rating
+                         / self.distance_from_city_center)
+        return round(washing_price, 1)
```
maxkatkalov, line 52: It can be implemented in one line return, it is more efficient. Try to make calculations in `return round()`

maxkatkalov (approved):
niiiice

Kroll410 (approved):

PR #689, opened 2023-08-06
Solution

denys-source:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center: float = distance_from_city_center
+        self.clean_power: int = clean_power
+        self.average_rating: float = average_rating
+        self.count_of_ratings: int = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if self.clean_power > car.clean_mark:
```
OlegatorLE, line 20: you are duplicating the code! method serve_cars has already done this check

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+            price: float = (car.comfort_class
+                            * (self.clean_power - car.clean_mark)
+                            * self.average_rating
+                            / self.distance_from_city_center)
+            return round(price, 1)
+        return 0
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
OlegatorLE, line 29: you are duplicating the code! method serve_cars has already done this check

OlegatorLE (changes requested):
code duplication fixes

app/main.py
denys-source: i have done this on purpose in case these methods will be used alone. without this check return value can be negative

OlegatorLE: I can't agree with you. This is the same check and the method will not run if the condition is not passed. Please, give an example where the result can be negative

denys-source: for example:
```
audi = Car(comfort_class=4, clean_mark=9, brand='Audi')

wash_station = CarWashStation(
    distance_from_city_center=5,
    clean_power=6,  # clean_power is less then audi.clean_mark
    average_rating=3.5,
    count_of_ratings=6)

income = wash_station.calculate_washing_price(audi)

print(income)  # -8.4
```

OlegatorLE: so we do not call this method to calculate income according to the condition of the problem) I think, in the future, we will be explained the difference between private, protected and ordinary methods in classes. you need to use the serve_cars method even if there is only one car

OlegatorLE (approved):

PR #690, opened 2023-08-06
Solution

OlegatorLE:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return \
+            car.comfort_class \
+            * (self.clean_power - car.clean_mark) \
+            * self.average_rating \
+            / self.distance_from_city_center
+
+    def wash_single_car(self, car: Car) -> None:
+        print(type(car))
```
innyshka, line 42: I think this line should be removed

innyshka (changes requested):
in general, everything is fine, just remove line 42

app/main.py
OlegatorLE: deleted

innyshka (approved):

Kroll410 (approved):

PR #692, opened 2023-08-06
Solution for py-car-wash-station

Barsh4ec:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+
+    def serve_cars(self, cars: list) -> int | float:
+        income = 0
+        for car in cars:
+            if self.clean_power >= car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> int | float:
```
ORENYT, line 30: int | float -> float

ORENYT, line 30: calculate_washing_price will always produce float, int annotation is wrong.

ORENYT (approved):
func calculate_washing_price will always produce float. int annotation is wrong. Everything else is correct. 
Value used to count price is float, so outcome will always be float according to math operations.

ORENYT (approved):
Well done in my opinion

Kroll410 (approved):

PR #696, opened 2023-08-07
Solution

9rosLove:


app/main.py
```diff
@@ -1,8 +1,57 @@
...
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
```
Kroll410, line 12: Apply formatter, you can try black here

Kroll410 (changes requested):

app/main.py
9rosLove: > still not fixed indentation

Can you add more details, pls?
In my opinion, the return type ") -> None:" should be either at the end of the 5th line or on the 6th line in alignment with "def".
But in the first case occurs  flake8 ".\main.py:2:80: E501 line too long (80 > 79 characters)" 
and in the second ".\main.py:3:5: E124 closing bracket does not match visual indentation".

9rosLove: or should I place each __init__ parameter from a new line?

app/main.py
```diff
@@ -1,8 +1,48 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str
+                 ) -> None:
```
IvanRamyk, line 3:
```suggestion
    def __init__(
        self, 
        comfort_class: int, 
        clean_mark: int, 
        brand: str
     ) -> None:
```

IvanRamyk: I believe this is the best way to deal with many argument in the function/method declaration

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,8 +1,50 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
IvanRamyk, line 2: style is still not fixed :) 
tests are failed because of that

make sure flake check is passed

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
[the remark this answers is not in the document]
IvanRamyk: if you have questions about how to do it, contact mentors in Slack

Kroll410 (approved):

PR #701, opened 2023-08-07
py-car-wash-station

yaroslav-mazurkevych:


app/main.py
```diff
@@ -1,8 +1,56 @@
...
+        return income
+
+    def calculate_washing_price(self, cars: Any) -> float:
+
+        income = cars.comfort_class * (self.clean_power - cars.clean_mark)\
+            * self.average_rating / self.distance_from_city_center
+
+        return round(income, 1)
+
+    def check_clean_mark_lower_then_power(self, car: Any) -> None:
```
Ihor-MA, line 44: Better to use car: Car. It shows that car expect object of type Car

Ihor-MA (approved):

app/main.py
yaroslav-mazurkevych: when i try to fix i have an error :N803 argument name 'Car' should be lowercase

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+
 class Car:
-    # write your code here
-    pass
+
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str) -> None:
```
IvanRamyk, line 10:
```suggestion
    self,
    comfort_class: int,
    clean_mark: int,
    brand: str
) -> None:
```

IvanRamyk: This way looks better :)

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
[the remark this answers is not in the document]
IvanRamyk: still not fixed

app/main.py
IvanRamyk: Car as a type annotation, instead of Any

IvanRamyk (changes requested):

Y-Havryliv (approved):
Nice!

PR #713, opened 2023-08-12
Solution

prochigor:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> int:
+        inc = 0
+        for car in car_list:
+            inc += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return inc
+
+    def calculate_washing_price(self, mark: Any) -> int | float:
```
LiudmylaKulinchenko: You should not use Any if you can specify type more clear
Change throughout the code

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> int:
+        inc = 0
```
LiudmylaKulinchenko: please, use descriptive name instead of `inc`
It could be a few more symbols but much higher readability

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: int | float,
```
wQuelS: this attribute can be `from 1.0 to 10.0` - so it is always a float. No int is gonna be stored here

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
```
wQuelS: why? what was the idea behind this decision ?

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> int:
+        income = 0
+        for car in car_list:
+            income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, mark: Car) -> int | float:
```
wQuelS: `mark` is not that good variable name for Car instance. It could be just `car` or `model`:) Fix everywhere

Also, return type is wrong, because you have division operator in your formula as well as rounding to one decimal place. So result will always have decimal part, even if it is `0`.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        if mark.clean_mark >= self.clean_power:
+            return 0
```
wQuelS: I guess this check should be implemented in `serve_cars()` method. That is because task description says, that washing can be done for cars with `clean_mark < clean_power`. No need to enter method before this check, because it can be ended like here in your approach - right after function call

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        price = mark.comfort_class * (self.clean_power - mark.clean_mark)
+        price = price * self.average_rating / self.distance_from_city_center
```
wQuelS: more pythonic way is to use `*=`  here.

Also, consider dividing this hard-to-read formula into smaller chunks with descriptive variable names. So it will be easy to understand what is going not only for you)

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> int:
```
wQuelS: are you sure in this return type ?

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            return 0
+        price = mark.comfort_class * (self.clean_power - mark.clean_mark)
+        price = price * self.average_rating / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, mark: Car) -> None:
+        if mark.clean_mark < self.clean_power:
+            mark.clean_mark = self.clean_power
+
+    def rate_service(self, evaluation: int) -> None:
```
wQuelS: Now we've reached the place where `mark` as a variable name is perfect. Use it instead of `evaluation` :)

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        rating = (self.average_rating * self.count_of_ratings + evaluation)
```
wQuelS: do you need those parentheses?

wQuelS (changes requested):

app/main.py
prochigor: I see, thank

app/main.py
prochigor: Yes, this was written in readme.md, clear

app/main.py
prochigor: I don't, there is only float

app/main.py
prochigor: Done

app/main.py
prochigor: Honestly, in the beginning, I had other solution, parentheses stay from the old one, I edit

app/main.py
prochigor: In "readme.md" used to be written "average_rating - average rating of the station, from 1.0 to 5.0, rounded to 1 decimal" and for some reason I decided to initialize it in the init method, It was a mistake)

app/main.py
prochigor: Yes, clear

LiudmylaKulinchenko (approved):
Good job! 🎈😸

PR #716, opened 2023-08-13
Solution

obohatov:


app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
Kroll410, line 2: Apply formatter, is it done in [example](https://mateacademy.slack.com/archives/C05HCQB4HQA/p1691763466046359)

Kroll410: You can use black for doing this easier and faster

Kroll410 (changes requested):

Kroll410 (approved):

PR #727, opened 2023-08-19
Solution

Daniil-Pankieiev:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
LiudmylaKulinchenko: Better to split parameters onto separate lines, if your line is too long

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.distance_from_city_center = distance_from_city_center
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = round(
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * (self.average_rating / self.distance_from_city_center)), 1)
+        return price
+
+    def wash_single_car(self, car: Car) -> bool:
+        return car.clean_mark < self.clean_power
```
LiudmylaKulinchenko: You should wash car here, not just make a check

LiudmylaKulinchenko: Also no need to use return here

LiudmylaKulinchenko (changes requested):

app/main.py
Daniil-Pankieiev: Fixed

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
+        result = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                result += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return result
```
Familenko: define that this is "income of CarWashStation" instead of "result"

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def rate_service(self, rate: float) -> None:
+        sum_rate = (self.average_rating * self.count_of_ratings) + rate
+        self.average_rating = round(
+            (sum_rate / (self.count_of_ratings + 1)), 1)
```
Familenko: as per best practice bracket need to locate on the same level         
self.average_rating = round(
            (sum_rate / (self.count_of_ratings + 1)), 1
)

I receive this advice from mentor before

Familenko (changes requested):

Naz-iv (commented):
well done

Naz-iv (approved):
all corrections applied

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                income_of_wash_station += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income_of_wash_station
+
+    def rate_service(self, rate: float) -> None:
+        sum_rate = (self.average_rating * self.count_of_ratings) + rate
+        self.average_rating = round(
+            (sum_rate / (self.count_of_ratings + 1)), 1
+        )
+        self.count_of_ratings += 1
```
wQuelS: 1) it seems like mark can be `int` only, for example, `3`, or `5`

2) maybe instead of using `self.count_of_ratings + 1` in formula and then `self.count_of_ratings += 1` after computation, you may increase the quantity of marks before formula and then calculate average mark, what do you think?

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = round(
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * (self.average_rating / self.distance_from_city_center)), 1)
+        return price
+
+    def wash_single_car(self, car: Car) -> bool:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
```
wQuelS: bool as return type? Are u sure?)

app/main.py
```diff
@@ -1,8 +1,49 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: int,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
```
Are u sure about some type annotations here ?

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.count_of_ratings = count_of_ratings
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.distance_from_city_center = distance_from_city_center
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = round(
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * (self.average_rating / self.distance_from_city_center)), 1)
```
wQuelS: please, be consistent in your formatting style. In the last function you have one style, here is another)
P.S. Formatting style in the `rate_service` func i prefer better, but it is up to you which one to choose. The only rule: it should be applied everywhere in your code:)

wQuelS (changes requested):

app/main.py
Daniil-Pankieiev: I apologize for my type annotations. I don’t know what happened to me then)

app/main.py
Daniil-Pankieiev: 1. I don’t fully understand. The annotations also indicate int.
2. Thank you, it really looks better

wQuelS: image
?

Daniil-Pankieiev: Fixed

wQuelS (approved):
well done

PR #729, opened 2023-08-20
Solution

SerhiiBudko:


tests/test_main.py
```diff
@@ -80,7 +80,7 @@ def test_car_is_washed(cars, wash_station, cars_clean_marks):
     ],
 )
 def test_car_cost_check_not_washed(car, wash_station, mark):
-    wash_station.calculate_washing_price(car)
+    wash_station.calculate_washing_price()
```
Dimosphen1, line 83: You should not modify tests, delete this change from your PR.

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self, distance_from_city_center: int | float,
```
Dimosphen1, line 13: You should place each argument on a new line, including `self`. It should be no line with two arguments here.

app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
Dimosphen1, line 2: You should follow one code style for your project, place `self` on a new line, and write `__init__` as in `CarWashStation`.

Dimosphen1 (changes requested):

SerhiiBudko:
Good day! Dmytro! I already changed the code by all yours recommendations! I have Error: “CarWashStation.calculate_washing_price() missing 1 required positional argument “Car”. I checked the code, "Car" argument is present, but the error is still the same!

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark)
+        calc_price = price * self.average_rating
+        calculate_price = round(calc_price / self.distance_from_city_center, 1)
```
Dimosphen1, line 37: You could make all calculations within one variable and write them, prefer this approach for the future taking into account requirements from the checklist instead of creating additional intermittent variables. After that, you could apply `round` to that single variable.

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return calculate_price
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, single_rate: float) -> None:
+        rating = self.average_rating * self.count_of_ratings + single_rate
+        total_ratings = self.count_of_ratings + 1
+        self.average_rating = round(rating / total_ratings, 1)
```
Dimosphen1, line 47: You could make all calculations within one variable and write them, prefer this approach for the future taking into account requirements from the checklist instead of creating additional intermittent variables. After that, you could apply `round` to that single variable.

Dimosphen1 (approved):
LGTM, several comments are left for consideration.

PR #734, opened 2023-08-23
Create 2 classes and many method

YaroslavYarovoy:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
```
Polyakiv-Andrey, line 12: After opening the parentheses, write all the arguments on a new line, including the first one, the closing bracket is also written on a new line

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return all_cost
+
+    def calculate_washing_price(self, car: Car) -> int:
+        if car.clean_mark <= self.clean_power:
+            cost_of_washing = round(
+                car.comfort_class * (self.clean_power - car.clean_mark)
+                * self.average_rating / self.distance_from_city_center, 1
+            )
+        else:
+            cost_of_washing = 0
```
Polyakiv-Andrey, line 37: instead of else declare Cost_of_washing in the first line of the method

Polyakiv-Andrey (changes requested):

app/main.py
YaroslavYarovoy: Can you explain why this is better? Few words pls)

PR #749, opened 2023-08-30
Solution

narizhiy:


app/main.py
```diff
@@ -1,8 +1,63 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
    def __init__(self,
                 comfort_class: int,
                 clean_mark: int,
                 brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,63 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int
+                 ) -> None:
```
```suggestion
    def __init__(
                 self,
                 distance_from_city_center: int,
                 clean_power: int,
                 average_rating: int,
                 count_of_ratings: int
                 ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,63 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```

Bohdan-Salamakha (changes requested):

app/main.py
narizhiy: Done

Bohdan-Salamakha (approved):

Bohdan-Salamakha (approved):

Bohdan-Salamakha (approved):

PR #755, opened 2023-09-05
'Solution'

nsdNite:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark <= self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> None:
+        old_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = (rating + old_rating) / self.count_of_ratings
```
Dimosphen1, line 50: You could write this intermittent result not in the instance attribute but rather in the local variable. Prefer this approach in the future.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #759, opened 2023-09-07
Solution

Andru37:


app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
    def __init__(self,
                 comfort_class: int,
                 clean_mark: int,
                 brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
```suggestion
    def __init__(self,
                 distance_from_city_center: int,
                 clean_power: int,
                 average_rating: float,
                 count_of_ratings: int) -> None:
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clear_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_of_clients: list) -> float:
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def rate_service(self, rate: int) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
+                                     * (self.count_of_ratings - 1) + rate)
+                                    / self.count_of_ratings, 1)
+        return
+
+    def wash_single_car(self, car: Car) -> bool:
```
Bohdan-Salamakha, line 42: bool typehint should be Optional in this case

Bohdan-Salamakha (changes requested):

app/main.py
Andru37: I realized that typehint with bool is not necessary.

app/main.py
Andru37: The argument writingscheme is clear

app/main.py
Andru37: The argument writing scheme is clear

app/main.py
Andru37: Made the changes using the example

app/main.py
Andru37: But when deleting a typehint bool, the check flaske8 fails

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_of_clients: list[Car]) -> float:
+        income = 0
+        for car1 in cars_of_clients:
+            if self.wash_single_car(car1):
+                car = car1
+                income += self.calculate_washing_price(car1)
+                car.clean_mark = self.clear_power
```
olena-hul: The variable `car1` is redundant. You can iterate using just `car` name

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clear_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def rate_service(self, rate: int) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
+                                     * (self.count_of_ratings - 1) + rate)
+                                    / self.count_of_ratings, 1)
+        return
```
olena-hul: this return statement does nothing, so it can be removed

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def wash_single_car(self, car: Car) -> bool:
+        if self.clear_power > car.clean_mark:
+            return True
```
olena-hul:
```suggestion
        return self.clear_power > car.clean_mark:
```

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #765, opened 2023-09-11
py_car_wash_station Solution

Vladkrit11:


app/main.py
```diff
@@ -1,8 +1,47 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
     ) -> None:
```

pavlejviki: Please follow this formatting throughout your code

pavlejviki (changes requested):

app/main.py
Vladkrit11: Made it in the next commit

pavlejviki (approved):
Good job!

PR #769, opened 2023-09-13
Solution

maxvell4:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self, distance_from_city_center: int,
+            clean_power: int,
+            average_rating: int,
```
viktoria-rybenchuk, line 15: Perhaps it would be better to use floating-point numbers when referring to average ratings

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center, 1)
+
```
viktoria-rybenchuk, line 28: you can format the code for better readability by placing the closing parenthesis on the same line as the round function call. 
return round(
    car.comfort_class * (self.clean_power - car.clean_mark)
    * self.average_rating / self.distance_from_city_center, 1
)

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #775, opened 2023-09-18
'Solution'

Chacer-Rider:


app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self: None, comfort_class: None,
```
Dimosphen1: Reformat your code so that all arguments in the `__init__` including `self` would start from a new line.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                 brand: None) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self: None, distance_from_city_center: None,
```
Dimosphen1: Reformat your code so that all arguments in the `__init__` including `self` would start from a new line.

app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self: None, comfort_class: None,
+                 clean_mark: None,
+                 brand: None) -> None:
```
Dimosphen1: Change your type annotation to indicate what types you expect to obtain for each argument, they should not be `None`.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self: None, distance_from_city_center: None,
+                 clean_power: None,
+                 average_rating: None,
+                 count_of_ratings: None) -> None:
```
Dimosphen1: Change your type annotation to indicate what types you expect to obtain for each argument, they should not be `None` here and in other places.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self: None, cars: None) -> None:
```
Dimosphen1: Do not use type annotation for `self` here and in other parts of your code. Indicate that you expect to obtain a list of `Car` instances here for `cars`.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return self.calculate_washing_price(
+            [car_to_wash for car_to_wash in cars
+             if car_to_wash.clean_mark < self.clean_power])
```
Dimosphen1, line 32: Your closing parenthesis `)` should be on a new line to match the style of the `(`.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+
+    def calculate_washing_price(self: None, cars: None) -> None:
+        result = 0
+        if type(cars) is list:
```
Dimosphen1, line 39: Consider that this method would be applicable to a single `Car` instance only as mentioned in the task, reformat your code to match that condition, and do not add redundant logic.

Dimosphen1 (changes requested):

app/main.py
viktoria-rybenchuk: Using None as the type annotation for self in a method signature is incorrect because it does not accurately represent the expected type. In Python, the first parameter of an instance method is conventionally named self, and its type is expected to be the class or instance itself.

viktoria-rybenchuk (changes requested):
Please read the comments I have provided

PR #781, opened 2023-09-23
Solution

OLENA-KALITSINSKA:


app/main.py
```diff
@@ -1,8 +1,48 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
    def __init__(self,
                 comfort_class: int,
                 clean_mark: int,
                 brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
```suggestion
    def __init__(self,
                 distance_from_city_center: float,
                 clean_power: int,
                 average_rating: float,
                 count_of_ratings: int) -> None:
```

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
```

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> None:
+        self.average_rating = ((self.average_rating
+                                * self.count_of_ratings
+                                + rating)
+                               / (self.count_of_ratings + 1))
```
viktoria-rybenchuk: Instead of writing one long formula in code, it is better to break it into several parts, while giving them understandable names.

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Well done!

PR #793, opened 2023-09-29
Car Wash Station Solution.

Lyutillis:


Naz-iv (approved):
looking good, one thing - return type for __init__ should be None since this method does not return anything

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_mark_diff = self.clean_power - car.clean_mark
+        divident = car.comfort_class * clean_mark_diff * self.average_rating
+        divider = self.distance_from_city_center
+        return round(divident / divider, 1)
+
+    def wash_single_car(self, car: Car) -> float:
```
fsocie7y: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.wash_single_car(car)
+
+        return income
```
fsocie7y: income should be rounded too

app/main.py
fsocie7y: No need to use return here

app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
fsocie7y, line 2: fix type annotation, init should return None

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                 brand: str) -> object:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
```
fsocie7y, line 12: also here

fsocie7y (changes requested):

app/main.py
Lyutillis: But I need to return the cost of washing the car. Should I separate methods wash_single_car and calculate_washing_price?

fsocie7y: yep, calculate_washing_price u can use in sevre_cars
and wash_single_car should return nothing

fsocie7y: image
and wash_single_car u should rewrite like this

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌺

PR #795, opened 2023-09-29
Solution

Naz-iv:


vshvanska (approved):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
wQuelS: if this method is called from elsewhere - logic will work not as it is intended.
A car should be wahsed only if `clean_power` is greater than `clean_mark`

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def rate_service(self, wash_rating: int) -> None:
```
wQuelS: wash_rating is not that good variable name here:)

wQuelS (changes requested):

app/main.py
Naz-iv: I had this check in previous rev. but decided to remove since it creates code duplication (checking same condition in both wash_single_car and serve_cars). So I removed this condition because I was thinking that car only gets washed if it's been served  by car wash station (called from serve_car method).

app/main.py
Naz-iv: noted, thanks

fsocie7y (approved):
GJ) 👍

PR #798, opened 2023-09-30
solution Washing Station

DenPrislipskyi:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                 count_of_ratings: int,
+                 ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_cars: list) -> float:
+        income = 0
+        for i in list_cars:
```
petrykivd: You can use more descriptive and correct variable names, like car.

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
```
petrykivd: Must be float : distance_from_city_center - how far station from the city center, from 1.0 to 10.0

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        for i in list_cars:
+            if len(list_cars) == 1:
+                income = self.wash_single_car(i)
+                break
+            if i.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(i)
+                i.clean_mark = self.clean_power
+        return income
+
+    def wash_single_car(self, single_car: Car) -> object | int:
```
petrykivd: This method should only wash the car, why u return here price for washing?

petrykivd (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str,
+                 ) -> None:
```
LiudmylaKulinchenko: fix formatting

app/main.py
```diff
@@ -1,8 +1,49 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int,
+                 ) -> None:
```
LiudmylaKulinchenko: fix formatting

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def wash_single_car(self, single_car: Car) -> None:
+        if single_car.clean_mark < self.clean_power:
+            single_car.clean_mark = self.clean_power
+
+    def calculate_washing_price(self, car: Car) -> float:
+        part_1 = car.comfort_class * (self.clean_power - car.clean_mark)
+        part_2 = self.average_rating / self.distance_from_city_center
+        total = part_1 * part_2
+        return round(total, 1)
```
LiudmylaKulinchenko: use descriptive names
And you can do multiplicatoin and rounding in the return
Or just use parentheses instead of creating variables

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def rate_service(self, rating_number: float) -> object:
+        sum_old_points = self.average_rating * self.count_of_ratings
+        sum_new_points = sum_old_points + rating_number
+        self.count_of_ratings += 1
+        new_average_rating = sum_new_points / self.count_of_ratings
+        self.average_rating = round(new_average_rating, 1)
+        return self.average_rating
```
LiudmylaKulinchenko: try to use much less variables, or just use parentheses

LiudmylaKulinchenko: no need to return anything here

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        if single_car.clean_mark < self.clean_power:
+            single_car.clean_mark = self.clean_power
+
+    def calculate_washing_price(self, car: Car) -> float:
+        part_1 = car.comfort_class * (self.clean_power - car.clean_mark)
+        part_2 = self.average_rating / self.distance_from_city_center
+        total = part_1 * part_2
+        return round(total, 1)
+
+    def rate_service(self, rating_number: float) -> object:
```
LiudmylaKulinchenko: You should use more detailed annotations instead `object`. In this case here should be None

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (commented):
Please do not mark conversations as resolved so we could see them when reviewing next time

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_cars: list) -> float:
+        income = 0
+        for car in list_cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
```
LiudmylaKulinchenko: You don't need it here, you do this action in wash_single_car method

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def wash_single_car(self, single_car: Car) -> None:
```
LiudmylaKulinchenko: You can name it `car` to make more correctly. Because in this case _single_ means that the method washes only one car by the time, not some specific single cars (sounds like `single car` is another type of car)

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice work! ☀🌺🌼

PR #802, opened 2023-10-01
Dev

neagarr:


app/main.py
```diff
@@ -1,8 +1,58 @@
...
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+            , 1)
+        return washing_cost
+
+    def wash_single_car(self, car: Car) -> Car:
+        car.clean_mark = self.clean_power
+        return car
```
vsmutok: image

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        total_cost = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total_cost += CarWashStation.calculate_washing_price(self, car)
+                CarWashStation.wash_single_car(self, car)
```
```suggestion
                total_cost += self.calculate_washing_price(self, car)
               self.wash_single_car(self, car)
```

vsmutok (changes requested):

app/main.py
neagarr: OK. But after all, then we will get a repetition of code, because the cars already sorted by the given condition in the another method ("serve_cars")

app/main.py
neagarr: ok

vasylhnatiuk (approved):

PR #806, opened 2023-10-01
Solution

MiskoRuslan:


D1st3f (approved):
Can be shorter, but all looks good !

app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
LiudmylaKulinchenko: fix formatting

app/main.py
```diff
@@ -1,8 +1,43 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
```
LiudmylaKulinchenko: fix formatting

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark > self.clean_power:
+                continue
+            income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class * (self.clean_power - car.clean_mark)
```
LiudmylaKulinchenko: `returns number rounded to 1 decimal` - fix that

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, response: float) -> None:
+        new_average_rating = self.average_rating * self.count_of_ratings
+        new_average_rating += response
+        self.count_of_ratings += 1
+        self.average_rating = round(
+            new_average_rating / self.count_of_ratings, 1)
```
LiudmylaKulinchenko: Bad formatting. Better do this:
round(
    ...,
    1
)

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class * (self.clean_power - car.clean_mark)
+                * self.average_rating / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, response: float) -> None:
+        new_average_rating = self.average_rating * self.count_of_ratings
+        new_average_rating += response
```
LiudmylaKulinchenko: Is it new avarage rating? Use self-descriptive names

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+                continue
+            income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class * (self.clean_power - car.clean_mark)
+                * self.average_rating / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> None:
```
LiudmylaKulinchenko, line 41: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
LiudmylaKulinchenko:
```suggestion
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) -> None:
```

fix formatting here

LiudmylaKulinchenko (changes requested):

app/main.py
MiskoRuslan: done 😎

fsocie7y (approved):
gj 🚀

PR #807, opened 2023-10-01
Solution

mykolamateichuk:


vshvanska (approved):

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        """
+        Calculates washing price depending on a comfort class of a car,
+        a difference between clean power of a washing station and a clean mark
+        of a car, and a rating to distance to center ratio
+        :param car:
+        :return:
+        """
+        difference_between_clean_power_and_clear_mark = (self.clean_power
```
LiudmylaKulinchenko: use shorter variable names

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        :return:
+        """
+        income = 0
+
+        for car in cars:
+            income += self.wash_single_car(car)
+
+        return income
+
+    def wash_single_car(self, car: Car) -> float:
```

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        """
+        Washes each car in the given list and returns the total income
+        :param cars:
+        :return:
+        """
```
LiudmylaKulinchenko: There is no need to add docstrings if the logic is pretty simple and understandable without it

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
```
LiudmylaKulinchenko, line 36: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice work! 🌼🌼☀

PR #809, opened 2023-10-02
Solution added!

D1st3f:


MiskoRuslan (approved):
I like everything, but on the 2nd and 10th lines, for readability, I would move the arguments to new lines

app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
LiudmylaKulinchenko: fix formatting. Split each parameter into one line
Fix everywhere

app/main.py
```diff
@@ -1,8 +1,38 @@
...
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class * abs(car.clean_mark - self.clean_power)
```
LiudmylaKulinchenko: `returns number rounded to 1 decimal` - fix that

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                * self.average_rating / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> Car:
```

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        car.clean_mark = self.clean_power
```

LiudmylaKulinchenko (changes requested):

RVChornyy (approved):

fsocie7y (approved):
GJ 👍

PR #810, opened 2023-10-02
Solution

Shatrovskyi:


Wonsky1 (approved):

app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
LiudmylaKulinchenko: It will be more readable if you split each parameter into one line

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: "Car") -> float:
```
LiudmylaKulinchenko: You can skip quotes in annotations here

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        washing_price = car.comfort_class * \
+            (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
```
LiudmylaKulinchenko: Please, don't use `\` separators, just add parentheses to split calculations into lines

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: "Car") -> int:
```
LiudmylaKulinchenko: No need to return anything here

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        car.clean_mark = self.clean_power
```

LiudmylaKulinchenko (changes requested):

app/main.py
Shatrovskyi: But It won't work there if I delete return-> income += self.calculate_washing_price(car)

LiudmylaKulinchenko: So you have to change your logic in another function. `wash_single_car` should still return nothing, like required in task description

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        washing_price = (
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center
+        )
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> int:
```
LiudmylaKulinchenko: You have to change your logic in another function. `wash_single_car` should still return nothing, like required in task description

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #811, opened 2023-10-02
Implement py-car-wash-station

SevKrok:


vshvanska (approved):

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+            income += self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * self.average_rating
+            * (self.clean_power - car.clean_mark)
+            / self.distance_from_city_center,
+            1)
```
LiudmylaKulinchenko: Bad formatting. Better do this:
round(
    ...,
    1
)

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+            price = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+
+        return price
+
+    def rate_service(self, mark: int) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + mark)
+            / (self.count_of_ratings + 1),
+            1)
```
LiudmylaKulinchenko: fix formatting also here

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * self.average_rating
+            * (self.clean_power - car.clean_mark)
+            / self.distance_from_city_center,
+            1)
+
+    def wash_single_car(self, car: Car) -> float:
```

LiudmylaKulinchenko (changes requested):

app/main.py
SevKrok: But why?
I think this better than use "if" two times and it's contains in one func wash single car

SevKrok: ok
understood

Sel-Fisher (approved):
as always the best and cleanest code i've ever seen

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                self.wash_single_car(car)
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * self.average_rating
+            * (self.clean_power - car.clean_mark)
+            / self.distance_from_city_center,
+            1)
```
LiudmylaKulinchenko:
```suggestion
        return round(
            car.comfort_class * self.average_rating
            * (self.clean_power - car.clean_mark)
            / self.distance_from_city_center, 1
        )
```

fix everywhere

LiudmylaKulinchenko: You can find it in the previous task [checklist](https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md) (at the end of it)

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done 🚀

PR #814, opened 2023-10-02
Solution

skoval7:


RVChornyy (changes requested):
in methods: calculate_washing_price and wash_single_car there is an attribute "car". The discription for this atribute must not be "str", because attribute "car" is not a string, but class instance

Shatrovskyi:
Agree with RVChornyy. Additionally from my perspective, we don't necessarily use the "if " statement in  wash_single_car() cause we check the same in serve_cars().

Shatrovskyi (approved):

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
LiudmylaKulinchenko: Split each parameter into line to make it more readable

app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
+        income = 0.0
+        for car in car_list:
+            if car.clean_mark < self.clean_power:
+                wash_cost = self.calculate_washing_price(car)
+                income += wash_cost
```
LiudmylaKulinchenko: remove `wash_cost` variable here, it is not needed

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+    def wash_single_car(self, car: list) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> float:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rating)
+            / (self.count_of_ratings + 1),
+            1  # Round to one decimal place
```
LiudmylaKulinchenko: remove comment

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: str) -> None:
+        price = (
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center
+        )
+        return round(price, 1)
+
+    def wash_single_car(self, car: list) -> None:
```
LiudmylaKulinchenko: fix `car` annotation

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+    def serve_cars(self, car_list: list) -> float:
+        income = 0.0
+        for car in car_list:
+            if car.clean_mark < self.clean_power:
+                wash_cost = self.calculate_washing_price(car)
+                income += wash_cost
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: str) -> None:
```
LiudmylaKulinchenko: same here about `car` annotation

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
```
LiudmylaKulinchenko, line 11: you can also split `self` into another line

LiudmylaKulinchenko (approved):
Well done! 🔥

PR #845, opened 2023-10-12
Solution

CrazyHilly:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
+
+    def can_be_cleaned(self, mark: int) -> bool:
+        if mark > self.clean_mark:
+            return True
```
ArsenPidhoretskyi:
```suggestion
        return mark > self.clean_mark
```

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+
+    def update_mark(self, mark: int) -> None:
+        if mark > self.clean_mark:
```
ArsenPidhoretskyi: In this case you can reuse can_be_cleaned

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
+
+    def can_be_cleaned(self, mark: int) -> bool:
+        return mark > self.clean_mark
+
+    def update_mark(self, mark: int) -> None:
+        if self.can_be_cleaned:
```
NikOneZ1: You should call the can_be_cleaned function to get the result
```suggestion
        if self.can_be_cleaned(mark):
```

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #859, opened 2023-10-21
Done

Moonralex:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def serve_cars(self, car_list: list) -> int | float:
+        income = 0
+        for car in car_list:
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: "Car") -> int | float:
```
```suggestion
    def calculate_washing_price(self, car: Car) -> int | float:
```

pavlejviki: don't use quotes in type annotations

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        if result <= 0:
+            return 0
+        return result
+
+    def wash_single_car(self, car: "Car") -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int | float) -> None:
+        all_stars = self.count_of_ratings * self.average_rating + rating
```
```suggestion
        total_rating = self.count_of_ratings * self.average_rating + rating
```

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+    def serve_cars(self, car_list: list) -> int | float:
+        income = 0
+        for car in car_list:
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: "Car") -> int | float:
+        washed = self.clean_power - car.clean_mark
```
```suggestion
        clean_power_difference = self.clean_power - car.clean_mark
```

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        how_expensive = self.average_rating / self.distance_from_city_center
```
```suggestion
        rating_to_distance_ratio = self.average_rating / self.distance_from_city_center
```

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        return result
+
+    def wash_single_car(self, car: "Car") -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int | float) -> None:
+        all_stars = self.count_of_ratings * self.average_rating + rating
+        self.average_rating = round(all_stars / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
pavlejviki, line 59: For improved readability and understanding, variables should have descriptive names that explain their purpose or the data they hold. I gave some suggestions

pavlejviki (changes requested):

app/main.py
Moonralex: TY, it looks better

app/main.py
Moonralex: I wrote exactly that way, because in the tasks on the site I keep an error flake8(F821) when I write without quotes.

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+        rating_to_distance_ratio = (self.average_rating
+                                    / self.distance_from_city_center)
+        result = round(
+            car.comfort_class
+            * clean_power_difference
+            * rating_to_distance_ratio,
+            1
+        )
+        if result <= 0:
+            return 0
```
viktoria-rybenchuk: why do you need this check?

viktoria-rybenchuk (changes requested):

app/main.py
Moonralex: When we use the calculate_washing_price method with a car whose clean_mark is greater than clean_power, we get a negative value. The logic of this method is to calculate the price for washing the car, not to pay extra for overly clean cars.

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> int | float:
```
Dimosphen1, line 26: You could indicate that you expect to obtain a list of `Car` instances for `car_list` in the type annotation.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #865, opened 2023-10-24
Solution

eLQeR:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        diffence = self.clean_power - car.clean_mark
+        total = car.comfort_class * diffence * self.average_rating
+        income = total / self.distance_from_city_center
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
```
ArsenPidhoretskyi: provide inner type for list

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        sum_of_incomes = 0
+        cars_for_washing = [
+            car
+            for car in cars
+            if car.clean_mark < self.clean_power
+        ]
+        for car in cars_for_washing:
+            sum_of_incomes += self.calculate_washing_price(car)
```
ArsenPidhoretskyi: Just add if statement here instead of cars_for_washing comprehension

ArsenPidhoretskyi (changes requested):

app/main.py
eLQeR: I dont get what inner type means, like list[Optional : Car] or what? Thanks

ArsenPidhoretskyi: list[Car]

ArsenPidhoretskyi (approved):
Well done!

PR #868, opened 2023-10-25
'Solution'

aleksandra-prognimak:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        incomes = []
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                incomes.append(self.calculate_washing_price(car))
+                self.wash_single_car(car)
+
+        return sum(incomes)
```
LiudmylaKulinchenko: Skip creating list here and using sum. You can just add incomes in the loop

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great ☀

PR #870, opened 2023-10-26
Solution

Marinel444:


app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
```
pavlejviki, line 10:
```suggestion
def __init__(
         self,
         comfort_class: int,
         clean_mark: int,
         brand: str
 ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return (car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> float:
+        if self.check_clean_mark(car):
+            price = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            print(price)
```
pavlejviki: remove print

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        price = [self.wash_single_car(car) for car in cars]
+        return round(sum(price), 1)
```
pavlejviki: instead of creating a list with list comprehension,you can  directly generate and sum up the values using a generator expression

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+            car.clean_mark = self.clean_power
+            print(price)
+            return round(price, 1)
+        return 0
+
+    def rate_service(self, grade: int) -> None:
+        new_grade = round(((self.average_rating
+                            * self.count_of_ratings)
+                           + grade)
+                          / (self.count_of_ratings + 1), 1)
```
pavlejviki: fix formatting

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return round(sum(price), 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> float:
+        if self.check_clean_mark(car):
```
pavlejviki: you can replace the method call with the comparison directly. there is no need to create a separate method for this logic

pavlejviki (changes requested):

app/main.py
Marinel444: what's wrong here? I don't see the message

pavlejviki (approved):
Well done!

PR #878, opened 2023-11-02
Solution

Molodoirockstar:


app/main.py
```diff
@@ -1,8 +1,57 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,57 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        filtered_cars = [
+            car
+            for car in cars
+            if car.clean_mark < self.clean_power
+        ]
+        income = 0
```
ArsenPidhoretskyi: No need to filter that just add this condition inside loop

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            income += self.calculate_washing_price(car)
```
ArsenPidhoretskyi: Due to task you must add condition:
if car.clean_mark >= self.clean_power:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #882, opened 2023-11-07
"car_wash_station"

380974261314kv:


app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
```
viktoria-rybenchuk: Move each parameter, including self, to a new line is a recommended practice for improving code readability,

viktoria-rybenchuk: fix in all places

viktoria-rybenchuk (changes requested):

app/main.py
380974261314kv: ok. code  has already fixed. please check it again

app/main.py
```diff
@@ -1,8 +1,50 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self,
```
pavlejviki: self should be on new line as well

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
                 brand: str
             ) -> None:
```

app/main.py

app/main.py
pavlejviki: closing parenthesis should be on new line as well

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                 * self.average_rating
+                 / self.distance_from_city_center)
+        return price if price > 0 else 0
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            income += self.calculate_washing_price(car)
+
+            if self.wash_single_car(car):
```
pavlejviki: wash_single_car method suggests it should return None , but you are using an if statement to check the result of self.wash_single_car(car) as if it was returning a boolean or some other value that can be evaluated in a conditional.

The method should either:
Return a boolean value (for example, True if the car was washed successfully, and False otherwise), in which case the type hint and the method implementation need to be changed.
Or not be used in a conditional check if it's only performing an action without determining success/failure and thus truly returns None.

pavlejviki (changes requested):

app/main.py
380974261314kv: regarding the code:
in the for loop there is an additional check to change the value of clean_mark for each machine.
image

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                car.clean_mark = self.clean_power
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, mark: int) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
```
Polyakiv-Andrey: After opening the parentheses, write each the arguments on a new line, including the first one

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self,
```
Polyakiv-Andrey: After opening the parentheses, write each the arguments on a new line, including the first one

app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self,
+                 comfort_class: int,
```
Polyakiv-Andrey: After opening the parentheses, write each the arguments on a new line, including the first one

Polyakiv-Andrey (changes requested):

app/main.py
viktoria-rybenchuk: not fixed

app/main.py
viktoria-rybenchuk: not fixed. You should check `if self.clean_power > car.clean_mark` and then call method `wash_single_car`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+            self.wash_single_car(car)
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> bool:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, mark: int) -> float:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
```
```suggestion
  self.average_rating = round(
            (self.average_rating
```

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #883, opened 2023-11-07
Solution

Maxboond:


app/main.py
```diff
@@ -1,8 +1,49 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
pavlejviki: self and the closing parenthesis should be from a new line as well

pavlejviki (changes requested):

app/main.py
```suggestion
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) -> None:
```

viktoria-rybenchuk: you should move each parameter on a new line. Fix in all places

viktoria-rybenchuk (changes requested):

app/main.py
Maxboond: so I have each parameter in a new line, or I don't understand something

Dimosphen1 (approved):
LGTM, well done!

app/main.py
Dimosphen1: Your first choice is also legit; however, when you use type annotation and, especially in the `__init__`, it is conventional to place each argument on a new line and leave opening and closing parenthesis without arguments on the same line.

PR #886, opened 2023-11-09
Solution

rinotokun:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
```
viktoria-rybenchuk: specify type annotation for `cars`

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+    def rate_service(self, wash_rating: int) -> None:
+        self.average_rating = (
+            (self.average_rating * self.count_of_ratings + wash_rating)
+            / (self.count_of_ratings + 1)
+        )
+        self.average_rating = round(self.average_rating, 1)
+        self.count_of_ratings += 1
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
```
viktoria-rybenchuk: You can improve the readability of your code by using parentheses for line continuation instead of backslashes.

viktoria-rybenchuk (changes requested):

app/main.py
rinotokun: But I already specified the annotation type for the car variable in the serv_cars function. The function accepts a list of cars "(self, cars: list)". Or did I misunderstand you?

viktoria-rybenchuk: You have already fixed it)

viktoria-rybenchuk (approved):
Great!

app/main.py
rinotokun: It dawned on me that I needed to clarify the type of objects in the list

PR #888, opened 2023-11-10
Solution

CANR-95:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        for car in cars:
+            if self.need_to_wash(car):
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        result = \
+            car.comfort_class * (self.clean_power - car.clean_mark) * \
+            self.average_rating / self.distance_from_city_center
```
EdAlekseiev: Use `()` instead of `\`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center
+        )
+        return round(result, 1)
+
+    def need_to_wash(self, car: Car) -> bool:
+        return self.clean_power > car.clean_mark
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
ArsenPidhoretskyi, line 44: In this method we also need to check if we need to wash car

ArsenPidhoretskyi (changes requested):

app/main.py
CANR-95: I have method  "need to wash", it check if we need to wash car
and i check it on line 28

YukihiroSM (approved):
Looks good to me!

PR #911, opened 2023-11-27
wash solution

OleksandrHubar:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
Dimosphen1: It is a conventional way to place all arguments including self on a new line in the `__init__` so that there would be no arguments on the same lines with opening and closing parenthesis. Rewrite `__init__` here and in other classes to encounter that.

tests/test_main.py
```diff
@@ -39,6 +39,7 @@ def test_car_wash_station(cars, wash_station, total_cost):
     income = wash_station.serve_cars(cars)
     assert income == total_cost, f"Income should equal to {total_cost}"
 
+
```
Dimosphen1: Remove a change from this file from the pull request.

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        self.average_rating = self.average_rating * self.count_of_ratings
+        self.average_rating += rating
+        self.average_rating /= (self.count_of_ratings + 1)
+        self.average_rating = round(self.average_rating, 1)
```
Dimosphen1: Make all operations to write the value to the `self.average_rating` at once to exclude the situation of writing the intermittent state there.

Dimosphen1 (changes requested):

app/main.py
OleksandrHubar: @Dimosphen1 +

tests/test_main.py
OleksandrHubar: @Dimosphen1 I'm not quite sure what needs to be adjusted.
Should I change file with tests?

app/main.py
OleksandrHubar: @Dimosphen1 I have a flake8 error when do it in one line

tests/test_main.py
viktoria-rybenchuk: No, you don't need any changes in the tests. For now, you have added a blank line; please rollback all changes in this file

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
```suggestion
    def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str
) -> None:
```

viktoria-rybenchuk: The following format is considered more readable

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,51 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
```
Dimosphen1: In case you do not have any arguments on the same line as the closing parenthesis, your opening parenthesis should not have any arguments on the same line as well. Reformat your code to encounter that and make the `__init__` style the same as for the `Car` class.

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        diff_mark = self.clean_power - car.clean_mark
+        washing_price = car.comfort_class * diff_mark * self.average_rating
+        washing_price /= self.distance_from_city_center
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rt: float) -> None:
```
Dimosphen1: Choose another argument name instead of `rt` to be more descriptive.

Dimosphen1 (changes requested):

app/main.py
OleksandrHubar: @Dimosphen1 I got, thanks for explanation)

viktoria-rybenchuk (approved):
Good job!

PR #921, opened 2023-12-08
Compled

Volokitin-Vitaliy:


app/main.py
```diff
@@ -1,8 +1,46 @@
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
viktoria-rybenchuk: place each parameter on a new line

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def serve_cars(self, cars: list) -> int:
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: list) -> int:
```
viktoria-rybenchuk: specify type annotation for `car`

viktoria-rybenchuk: it should return `float` as you round result

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        cost = (
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+        )
+        return round(cost, 1)
+
+    def wash_single_car(self, car: list) -> int:
```
viktoria-rybenchuk: wrong type annotation for `car`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
```suggestion
    def __init__(
            self,
            comfort_class: int,
```

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
                 brand: str
             ) -> None:
```

pavlejviki: please start each parameter from new line starting from self.
opening and closing parenthesis should be on new lines as well

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #925, opened 2023-12-10
Car was station classes

cyb1ove:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> int:
+        clean_difference = self.clean_power - car.clean_mark
+        price = car.comfort_class * clean_difference * self.average_rating
+        price /= self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> bool:
```
fsocie7y, line 26: No need to use return here, u can provide calculations in serve_cars method

fsocie7y (changes requested):

app/main.py
cyb1ove: I don`t understand. Readme says, that function should return calculation. And code example has return in this function.

fsocie7y (approved):
well done ^_^

PR #946, opened 2023-12-20
'Solution'

JohnnyWalker010:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
```
Serhii-Leonenko: fix indentation here,  
`count_of_ratings: int) -> None:`

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.distance = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car: list) -> float:
```
Serhii-Leonenko: Use more descriptive type annotations `list[Car]`

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        served_cars_list = []
+        income = 0
+        for item in car:
+            if item.clean_mark < self.clean_power:
+                served_cars_list.append(item)
+        for car in served_cars_list:
+            income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
```
Serhii-Leonenko: use only one loop here

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: classmethod) -> float:
```
Serhii-Leonenko: fix type annotation for the car

Serhii-Leonenko: the same in another similar case

Serhii-Leonenko (changes requested):

app/main.py
JohnnyWalker010: Not clear for me: can't pass flake8 test for any other way of indentation except the actual one.

app/main.py
JohnnyWalker010: Thanks!

app/main.py
Serhii-Leonenko: you have to follow one style for brackets wrapping

```suggestion
    def __init__(self,
                 distance_from_city_center: float,
                 clean_power: int,
                 average_rating: float,
                 count_of_ratings: int) -> None:
```

understood?)

Serhii-Leonenko (changes requested):

kabust (approved):

app/main.py
Serhii-Leonenko: What about this ?)

Serhii-Leonenko (approved):

PR #951, opened 2023-12-22
wash-your-car-first-try

Yevhen23tk:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+        self.clean_power = clean_power
+
+    def serve_cars(self, cars: list) -> float:
```
NikOneZ1: Specify the list type 
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

NikOneZ1 (changes requested):

app/main.py
Yevhen23tk: ok, accept

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return round((car.comfort_class
+                      * (self.clean_power - car.clean_mark)
+                      * self.average_rating
+                      / self.distance_from_city_center),
+                     1
+                     )
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
ArsenPidhoretskyi: wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark;
Condition does not exist

ArsenPidhoretskyi (changes requested):

app/main.py
Yevhen23tk: ok, accept

YukihiroSM (approved):
Good job!

PR #953, opened 2023-12-23
Solution

rzaliznyi:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: object) -> Any:
```
ArsenPidhoretskyi: car is not object, it is Car

app/main.py
```diff
@@ -1,8 +1,48 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: Any,
+                 average_rating: float, count_of_ratings: Any) -> Any:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
ArsenPidhoretskyi: list of what? provide inner type annotations

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+            price = (
+                car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center
+            )
+            return round(price, 1)
+        return 0
+
+    def wash_single_car(self, car: object) -> Any:
```
ArsenPidhoretskyi: car is not object, it is Car

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+            return car.clean_mark
```
ArsenPidhoretskyi: return is redundant

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+    def wash_single_car(self, car: Car) -> Any:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round(
+            (total_rating + rating) / self.count_of_ratings, 1)
+        return self.count_of_ratings, self.average_rating
```
ArsenPidhoretskyi: This return is redundant

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            price = (
+                car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center
+            )
+            return round(price, 1)
+        return 0
+
+    def wash_single_car(self, car: Car) -> Any:
```
ArsenPidhoretskyi: It returns None

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> Any:
```
ArsenPidhoretskyi: It returns float

ArsenPidhoretskyi (changes requested):
Do not use Any as a type annotation because your code does not work with Any type. If I enter some string value your code will stop working but according your type annotations it should work

ArsenPidhoretskyi (approved):
Well done!

PR #972, opened 2024-01-05
Solution

PythonZem:


barabarinov (approved):

app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
Serhii-Leonenko:
```suggestion
    def __init__(
         self,
         comfort_class: int,
         clean_mark: int,
         brand: str
) -> None:
```
keep the one indent style

Serhii-Leonenko: the same in other similar cases

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int
+                 ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> float:
```
Serhii-Leonenko: Use more descriptive type annotations 

```suggestion
    def serve_cars(self, cars_list: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        return result
+
+    def calculate_washing_price(self, car: Car) -> float:
+        different = self.clean_power - car.clean_mark
+        rate_to_dist = self.average_rating / self.distance_from_city_center
+        price = car.comfort_class * different * rate_to_dist
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Serhii-Leonenko: If you call this method separately, it won't work properly, reread the task

Serhii-Leonenko (changes requested):

app/main.py
PythonZem: Thank you. I remember
Done!

app/main.py
PythonZem: Done!

app/main.py
PythonZem: I added a comparison in this method. Now it can work independently

Y-Havryliv (approved):
gj!

PR #974, opened 2024-01-07
Solution

isikach:


app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
barabarinov, line 5: [All arguments should be at new line](https://github.com/mate-academy/py-car-wash-station/pull/972#discussion_r1444226491). The same in other similar cases

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def wash_single_car(self, car: Car) -> int:
+        return self.clean_power \
+            if self.clean_power >= car.clean_mark else 0
+
+    def rate_service(self, new_rate: int) -> float:
+        self.average_rating = round(((self.average_rating
+                                      * self.count_of_ratings
+                                      + new_rate)
+                                    / (self.count_of_ratings + 1)), 1)
```
barabarinov, line 44: The same here.

barabarinov (changes requested):

app/main.py
isikach: Thank you for your code review but I respectfully disagree.
You can check it in this link:
https://peps.python.org/pep-0008/#indentation
Look please in correct and wrong examles

NEkropo1 (approved):
LGTM!

PR #980, opened 2024-01-09
Solution (without in-class decorators)

MaksymNikulyak:
Have a repeating condition in the code lines 23 and 35:
"if car.clean_mark < self.clean_power:"
Therefore, an in-class decorator might be a nice thing to do.


EdAlekseiev (approved):
Good job

PR #981, opened 2024-01-10
Solution

MaxymChyncha:


VladKyianytsia (approved):
Correct!

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        total_income = 0
+
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return total_income
```
Y-Havryliv: it should be rounded to 1 decimal

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round((car.comfort_class
+                      * (self.clean_power - car.clean_mark)
+                      * self.average_rating
+                      / self.distance_from_city_center), 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return total_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round((car.comfort_class
```
Y-Havryliv: it's hard to read a little bit, probably it's better to create some additional variables

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -33,16 +33,21 @@ def serve_cars(self, cars: list[Car]) -> float:
...
-        return round((car.comfort_class
-                      * (self.clean_power - car.clean_mark)
-                      * self.average_rating
-                      / self.distance_from_city_center), 1)
+        clean_diff = self.clean_power - car.clean_mark
+        washing_price = (
+            car.comfort_class * clean_diff
+            * self.average_rating
+            / self.distance_from_city_center
+        )
```
NEkropo1: why all operands are not on new lines? Can't get the logic.

NEkropo1 (changes requested):

NEkropo1 (approved):
Good job!

PR #983, opened 2024-01-10
Car wash station Solution

shaiduchyk:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int,
+                 income: float = 0.0) -> None:
```
mdubyna: Class CarWashStation has to take only 4 arguments(Income is not needed) - but i like it)

mdubyna (approved):
Good job!

app/main.py
Y-Havryliv: yes, fix it

Y-Havryliv (changes requested):

NEkropo1 (approved):
LGTM!

PR #987, opened 2024-01-10
Solution

Bulleter06:


tskopenko (approved):
Code done according to the rules)

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        total_income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income = self.calculate_washing_price(car)
+                total_income += income
```
```suggestion
                total_income += self.calculate_washing_price(car)
```
you can do it without additional variable

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        total_ratings = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = (total_ratings + rating) / self.count_of_ratings
+        self.average_rating = round(self.average_rating, 1)
+
+# Example usage:
```
Y-Havryliv: why do you need it?

Y-Havryliv (changes requested):

app/main.py
Bulleter06: self.average_rating = (total_ratings + rating) / self.count_of_ratings : it calculates the new average rating.                                                                                                                                                        self.average_rating = round(self.average_rating, 1) : ensures that the average rating is presented with a consistent level of precision.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        total_ratings = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = (total_ratings + rating) / self.count_of_ratings
+        self.average_rating = round(self.average_rating, 1)
```
NEkropo1: that's right, but so complicated, consider either break it down in smaller functions, which will show, what done there, or rewright it, so flow will be right.
Also I don't think it's necessary to make 1 more row to only round the answer, here, consider:
```suggestion
    def rate_service(self, rating: float) -> None:
        total_ratings = self.average_rating * self.count_of_ratings
        self.count_of_ratings += 1
        self.average_rating = round((total_ratings + rating) / self.count_of_ratings, 1)
```
or something similar.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+
+bmw = Car(3, 3, "BMW")
+audi = Car(4, 9, "Audi")
+mercedes = Car(7, 1, "Mercedes")
```
NEkropo1: I guess we don't need these lines.

NEkropo1 (changes requested):

app/main.py
```diff
@@ -38,10 +38,5 @@ def wash_single_car(self, car: Car) -> None:
     def rate_service(self, rating: float) -> None:
```
NEkropo1: also rating could be either float or int, I think )

NEkropo1 (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #990, opened 2024-01-10
Solution

tskopenko:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * (self.average_rating / self.distance_from_city_center)),
+            1
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
NataliiaKorzh: тут повинна бути умова if self.clean_power > car.clean_mark перед цим рядком

NataliiaKorzh (approved):

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
```
Y-Havryliv, line 12: `Create class CarWashStation, its __init__ method takes and stores 4 arguments:`

app/main.py

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        self.average_rating = round(
+            ((self.average_rating * self.count_of_ratings) + rating)
+            / (self.count_of_ratings + 1), 1
+        )
```
NEkropo1: consider breaking down complex functions into smaller parts for clarity and readability.

NEkropo1 (changes requested):

app/main.py
```diff
@@ -41,6 +40,5 @@ def wash_single_car(self, car: Car) -> None:
     def rate_service(self, rating: float) -> None:
         self.average_rating = round(
             ((self.average_rating * self.count_of_ratings) + rating)
-            / (self.count_of_ratings + 1), 1
-        )
+            / (self.count_of_ratings + 1), 1)
         self.count_of_ratings += 1
```
NEkropo1: Actually previous approach was better)
I meant to break down whole expression in a smaller parts, like:
```
change_count_of_wheels(self, added_wheels) -> None:
new_wheels_per_side = self.wheel_per_side + added_wheels
self.wheels_count = new_wheels_per_side * self.sides
```
instead of :
```
change_count_of_wheels(self, added_wheels) -> None:
new_wheels_per_side = (self.wheel_per_side + added_wheels) * self.sides
```
It's redundant in exact example, but I wanted to show the flow.

NEkropo1 (changes requested):

NEkropo1 (approved):
Great!

PR #991, opened 2024-01-10
Solution

kstorozhenko:


motivated2 (approved):
Seems good to me!

app/main.py
```diff
@@ -1,8 +1,51 @@
...
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
+        if comfort_class > 7 or comfort_class < 1:
+            print("invalid class value")
```
Y-Havryliv: prints give you nothing, if you want to add some validation the exceptions would be better

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,8 +1,57 @@
...
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
+        try:
+            if comfort_class > 7 or comfort_class < 1:
+                raise ValueError("invalid class value")
+        except ValueError as class_value_error:
+            print(class_value_error)
```
NEkropo1: whats this?)

NEkropo1 (changes requested):

app/main.py
kstorozhenko: Yaroslav wrote "Y-Havryliv 18 hours ago
prints give you nothing, if you want to add some validation the exceptions would be better" that it is better to use exceptions, I googled how they work and tried to implement)

> whats this?)

Serhii-Leonenko (approved):
Well done!

PR #992, opened 2024-01-10
'Solution'

NataliiaKorzh:


Dobryk12 (approved):

app/main.py
```diff
@@ -1,8 +1,47 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
```suggestion
    def __init__(self, 
    comfort_class: int,
```
it would be more readable to move it to the next line

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
```
Y-Havryliv: same here

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark > self.clean_power:
```
Y-Havryliv: it would be more simple to use `if car.clean_mark < self.clean_power:` and then do all the logic, and you can omit `continue`

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #993, opened 2024-01-10
implemented class Car and class CarWashStation in main.py

motivated2:


Serejka21 (approved):

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        price = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price += (self.calculate_washing_price(car))
+                self.wash_single_car(car)
+        return price
```
Y-Havryliv: it should be rounded to 1 decimal

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center, 1
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```

Y-Havryliv (changes requested):

NEkropo1 (approved):
Good job!

PR #1006, opened 2024-01-14
Solution

nicksetrakov:


app/main.py
```diff
@@ -1,8 +1,36 @@
...
+                self.wash_single_car(car)
+        return total_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        diff_wash_car = self.clean_power - car.clean_mark
+        return round((car.comfort_class * diff_wash_car * self.average_rating
+                      / self.distance_from_city_center), 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
YuriiKindrat: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return total_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        diff_wash_car = self.clean_power - car.clean_mark
+        return round((car.comfort_class * diff_wash_car * self.average_rating
+                      / self.distance_from_city_center), 1)
```
YuriiKindrat: try to increase readability here

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
YuriiKindrat: it is better to place one arg per line here, for better readability

app/main.py
```diff
@@ -1,8 +1,36 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
YuriiKindrat: it is better to place similar number of args at the line for better readability

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM!

PR #1010, opened 2024-01-18
Solution

yaaszp:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
+
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
NazarSheremet: Need to add a check if wash_station.clean_power is greater than car.clean_mark according to the task requirements
image

NazarSheremet (changes requested):

app/main.py
yaaszp: fixed. added additional condition

YuriiSmolii (approved):
LGTM

PR #1018, opened 2024-01-21
Solution

vsupruniuk:


app/main.py
```diff
@@ -1,8 +1,57 @@
...
+                car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center
+            ),
+            1,
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #1028, opened 2024-01-25
All good?

Olevsky-Ivan:


app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
pavlejviki: add each parameter on a new line

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int,
```
pavlejviki: add each parameter on a new line

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int | float:
```
```suggestion
    def serve_cars(self, cars: list[Car]) -> int | float:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,53 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1, line 5: It is a common practice to place each argument in the `__init__` on a new line, so that opening and closing parenthesis would not have any arguments on the same line. Follow this approach in the future.

Dimosphen1 (approved):
LGTM, well done! One comment regarding the code style is left for consideration.

PR #1034, opened 2024-01-26
Solution

SysoevDmitro:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
```
EdAlekseiev: Specify inner type of list
```suggestion
    def serve_cars(self, cars: list[Car]) -> int:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
YukihiroSM, line 11: if you are separating function arguments you should place each parameter in one line, or at least do something like this: 
```suggestion
    def __init__(
                 self, distance_from_city_center: float, clean_power: int,
                 average_rating: float, count_of_ratings: int) -> None:
```

https://peps.python.org/pep-0008/#indentation (1st paragraph)

YukihiroSM (approved):
Else looks good, so consider the comment about indentation in future :)

PR #1055, opened 2024-02-02
solved task

Molchanov-Ihor:


pavlejviki (changes requested):
tests failed. fix flake8

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
+        money = []
```
viktoria-rybenchuk: Create money as `0` and then add cash on each iteration  is better for memory efficiency

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        result = round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center,
+            1
+        )
+        return result
+
+    def wash_single_car(self,
+                        car: Car) -> None:
+        if car.clean_mark < self.clean_power:
```
viktoria-rybenchuk: this code is redundant. You have already checked it in `serve_cars` function

app/main.py
```diff
@@ -1,8 +1,48 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
```suggestion
    def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str
) -> None:
```

viktoria-rybenchuk: place each parameter on a new line including self. It provides better readability

viktoria-rybenchuk (changes requested):

app/main.py
Molchanov-Ihor: fixed

app/main.py
Molchanov-Ihor: fixed

app/main.py
Molchanov-Ihor: fixed

viktoria-rybenchuk (approved):
Well done!

PR #1058, opened 2024-02-04
implemented classes for car and wash station

RomanHlodann:


app/main.py
```diff
@@ -1,8 +1,37 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
viktoria-rybenchuk: place each parameter on a new line including self. Leave parentheses one a new line as well

app/main.py
```diff
@@ -1,8 +1,37 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
viktoria-rybenchuk: the same here

app/main.py
```diff
@@ -1,8 +1,37 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        return round(sum([self.wash_single_car(car) for car in cars]), 1)
```
```suggestion
        return round(sum(self.wash_single_car(car) for car in cars), 1)
```

viktoria-rybenchuk: You can use generator instead of list comprehension for better memory performance

app/main.py
```diff
@@ -1,8 +1,37 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        if self.clean_power >= car.clean_mark:
+            washing_price = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return washing_price
+        return 0
```
viktoria-rybenchuk: In this function you should implement  just this logic `wash_single_car` - method, that washes a single car, so it should have `clean_mark` equals wash station's `clean_power`, if `wash_station.clean_power` is greater than `car.clean_mark`;
Logic with calculation washing price move to the `serve_cars` method

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1)
+            , 1)
```
Dimosphen1, line 46: In case an opening bracket does not have any arguments, your closing bracket should not have any arguments as well. Take this into account in the future.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #1061, opened 2024-02-05
Implement Washing-cost-calculation

anrozhna:


app/main.py
```diff
@@ -1,8 +1,55 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * (self.average_rating / self.distance_from_city_center),
+            1
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power \
+            if (self.clean_power > car.clean_mark) else car.clean_mark
```
viktoria-rybenchuk: no need make this checking. You have already checked this logic in `serve_cars`

viktoria-rybenchuk (changes requested):

app/main.py
anrozhna: Done

Dimosphen1 (approved):
LGTM, well done!

PR #1063, opened 2024-02-06
add task solution

EugeneSherepa:


app/main.py
```diff
@@ -1,8 +1,52 @@
...
+
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        wash_cost = round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center,
+            1)
```
LiudmylaKulinchenko: Fix indentation here, according to the best practices
https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md

The braces must be located in one of two options: open and start with the text or have a line break between the text. Fix throughout the code

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        return wash_cost
+
+    def wash_single_car(self, car: Car) -> None:
```
LiudmylaKulinchenko, line 45: You need to check if clean_power is greater than clean_mark here because if you run this method separately from others it will not work correctly

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice work 💪

PR #1074, opened 2024-02-12
Solution

AlexTsikhun:
        # works without condition, cause I check above

Should I add condition? (in task description menshion about it, but I checked wash ability above), this means, that it not necessary?
And indentation in __init__ parameters is ok?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -41,9 +41,7 @@ class CarWashStation:
         return round(cost, 1)
 
     def wash_single_car(self, car: Car) -> None:
-        # works without condition, cause I check above
-        if self.clean_power > car.clean_mark:
-            car.clean_mark = self.clean_power
+        car.clean_mark = self.clean_power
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        power_for_car = self.clean_power - car.clean_mark
+        cost = (car.comfort_class
+                * power_for_car
+                * self.average_rating
+                / self.distance_from_city_to_center)
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        # works without condition, cause I check above
+        if self.clean_power > car.clean_mark:
```
viktoria-rybenchuk: you can remove this condition as it is redundant logic

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Nice!

PR #1075, opened 2024-02-12
Solution

rosttv:
I did the task with Interpreter 3.8, so I had to import "List"

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,21 +2,25 @@ from typing import List
 
 
 class Car:
-    def __init__(self,
-                 comfort_class: int,
-                 clean_mark: int,
-                 brand: str) -> None:
+    def __init__(
+        self,
+        comfort_class: int,
+        clean_mark: int,
+        brand: str
+    ) -> None:
         self.comfort_class = comfort_class
         self.clean_mark = clean_mark
         self.brand = brand
 
 
 class CarWashStation:
-    def __init__(self,
-                 distance_from_city_center: int,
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,48 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1: Place each argument on a new line including self, so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this to `__init__` for `CarWashStation` class as well.

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1), 1)
```
Dimosphen1: In case there are no arguments on the same line as the opening parenthesis, the closing parenthesis should not have any arguments on the same line as well.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1089, opened 2024-02-19
add solution to washing car

vdzhyranov:


app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+                (self.average_rating
+                    * car.comfort_class
+                    * (self.clean_power - car.clean_mark))
+                / self.distance_from_city_center,
+                1)
```
vdzhyranov, line 32: this alignment looks shitty, but I cannot split logic in different vars and use them because don't see a logic with all of these operations

YuriiSmolii (approved):
LGTM

PR #1095, opened 2024-02-22
Solution

skeletronys:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, cars: Car) -> float:
+        return (cars.comfort_class * (self.clean_power - cars.clean_mark)
+                * self.average_rating / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
viktoria-rybenchuk: this code is redundant, you have already defined this logic in `serve_cars` method

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+
+    def serve_cars(self, car_list: list) -> float:
+        income = 0.0
+        for car in car_list:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, cars: Car) -> float:
```
viktoria-rybenchuk: write correct variable name, you are expecting to obtain single car not cars

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: float) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
```
viktoria-rybenchuk: specify the type annotation for car_list as a list containing elements of type `Car`

viktoria-rybenchuk (changes requested):

app/main.py
skeletronys: The Car annotation cannot be specified here, as a floating-point value is returned

app/main.py
```diff
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1: Place each argument on a new line including `self` so that there would not be any arguments on the same line with opening and closing parenthesis. It would improve the readability. Apply this change to the  `CarWashStation` class as well

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
     pass
```
Dimosphen1: Delete `pass`

app/main.py
Dimosphen1: You should specify the type annotation for argument `car_list` from indicating that you expect to obtain a list to expecting obtaining a list of `Car` class instances. It does not relate to the return type annotation. Besides that, it would be beneficial to change variable naming from `car_list` to `cars` to not indicate the type of the variable in its name without necessity.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #1104, opened 2024-02-25
Solution for Classes

anastasiiakalini:


app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
Dimosphen1: Place each argument on a new line including `self` in the `__init__` for better code style and readability. There should be no arguments on the same line with `(` and `)`. Make this change in the `CarWashStation` as well

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+        return all_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            diff = self.clean_power - car.clean_mark
+            average, dist = self.average_rating, self.distance_from_city_center
+            price = car.comfort_class * diff * average / dist
+            return round(price, 1)
+        else:
```
Dimosphen1: There is no need for `else` here, the code after `return` will not execute anyway. Remove redundant `else`

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+            self.wash_single_car(car)
+            all_income += income
+
+        return all_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            diff = self.clean_power - car.clean_mark
+            average, dist = self.average_rating, self.distance_from_city_center
+            price = car.comfort_class * diff * average / dist
```
Dimosphen1: Perform this operation in one expression using the formatting mentioned in the checklist.

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+            return 0.0
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        average, count = self.average_rating, self.count_of_ratings
+        self.average_rating = (average * count + rate) / (count + 1)
+        self.average_rating = round(self.average_rating, 1)
```
Dimosphen1: Perform this operation in one expression using the formatting mentioned in the checklist. Do not assign and reassign the value to the instance attribute several times within one function, rather declare a variable and after that assign it to the instance attribute at once

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        all_income = 0
+        for car in cars:
+            income = self.calculate_washing_price(car)
+            self.wash_single_car(car)
+            all_income += income
+
+        return all_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
```
viktoria-rybenchuk: you should make this checking in `seve_cars`, if this condition pass then call  `calculate_washing_price

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+            price = (car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center)
+            return round(price, 1)
+        return 0.0
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
```
viktoria-rybenchuk: no need make this checking here

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+
+        return all_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            price = (car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center)
+            return round(price, 1)
+        return 0.0
```
viktoria-rybenchuk: there is no need to return 0, you have already defined all_income as initial value

viktoria-rybenchuk (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1108, opened 2024-02-28
Develop

Luxiv:
Task is done, please check


NazarSheremet (changes requested):
Flake failed for a reason, that length of line more than 79 characters, Need to divide it for two line 
image

YuriiSmolii (approved):
Good job!

PR #1109, opened 2024-02-28
Solution

olvyshh:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: [Car]) -> float:
```
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

EdAlekseiev (changes requested):

app/main.py
olvyshh: after changing a waring appears "Builtin 'list' cannot be parameterized directly "
Screenshot from 2024-02-28 21-10-37

olvyshh: The cause of the problem may be old python interpritator version (3.8 i use)

EdAlekseiev: You can use List from `typing` package

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
EdAlekseiev: You can use List from typing package

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #1110, opened 2024-02-28
'Solution'

o-zinevych:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+            else:
+                continue
```
EdAlekseiev: this block is redundant

EdAlekseiev (changes requested):

app/main.py
o-zinevych: Thanks, I was thinking about that

EdAlekseiev (approved):

PR #1119, opened 2024-03-05
Solution

AlexNechval:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) -> None:  # noqa: E501
```
NazarSheremet: No need to turn off the linter for this line
```suggestion
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) -> None:
```

NazarSheremet (changes requested):
Minor changes

YuriiSmolii (approved):
LGTM

PR #1124, opened 2024-03-06
Task solved

Sydorovich:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        temp = temp * (self.clean_mark - car.clean_mark)
+        return round(temp / self.distance, 1)
+
+    def wash_single_car(self, car: Any) -> None:
+        if self.clean_mark <= car.clean_mark:
+            return
+        car.clean_mark = self.clean_mark
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating *= self.count_of_ratings
```
YuriiSmolii: Explain why do you need this multiplication.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.distance = distance_from_city_center
+        self.clean_mark = clean_mark
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Any) -> float:
+        if self.clean_mark <= car.clean_mark:
+            return 0.0
+        temp = car.comfort_class * self.average_rating
+        temp = temp * (self.clean_mark - car.clean_mark)
```
YuriiSmolii: This is bad code construction, simplify it.

YuriiSmolii (changes requested):
Minor change and explanation requested.

app/main.py
Sydorovich: Simplified it in new commit

app/main.py
Sydorovich: Simplified it in new commit

Sydorovich:
Pls recheck the task after the changes.

PR #1126, opened 2024-03-08
Solution

Bujhm99:


app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.dist_to_center, 1)
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                car.clean_mark = self.clean_power
```
Boyok1337: It have to work with function, cause you set value and then call func for set new value that equal to previous

Boyok1337 (changes requested):

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.dist_to_center, 1)
+
+    def serve_cars(self, cars: list) -> float:
+        return sum([self.wash_single_car(car) for car in cars])
+
+    def wash_single_car(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            res = self.calculate_washing_price(car)
```
YuriiKindrat: use full names for variables result - not res

app/main.py
```diff
@@ -1,8 +1,43 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        # how far station from the city center, from 1.0 to 10.0
+        self.dist_to_center = distance_from_city_center \
+            if distance_from_city_center != 0 else 0.00000000001
```
YuriiKindrat: why do u need this condition ? also try to avoid using linebreakers

YuriiKindrat (changes requested):
please, remove all unnecessary comments

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.dist_to_center, 1)
+
+    def serve_cars(self, cars: list) -> float:
+        return sum([self.wash_single_car(car) for car in cars])
```
LiudmylaKulinchenko: No need to use [] with `sum`

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
LGTM!

PR #1143, opened 2024-03-14
Solution

CharmNiovi:


app/main.py
```diff
@@ -1,8 +1,58 @@
 class Car:
-    # write your code here
-    pass
+    __slots__ = ("comfort_class", "clean_mark", "brand")
```
panicua, line 2: an interesting implementation, I didn't know about ```__slots__``` untill now, had to look for the info about it :)

panicua (approved):
GJ

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return (
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+        )
+
+    def wash_single_car(self, clean_mark: int) -> bool:
+        return clean_mark <= self.clean_power
```
LiudmylaKulinchenko: `wash_single_car - method, that washes a single car` - you should wash a car here

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ!

PR #1145, opened 2024-03-14
Solution

panicua:
Very interesting task. If I didn't miss anything in the task explanation there are no words about the rate_service function return. So it wasn't clear for me untill I found an error in tests. It should return rounded number to 1 declimal point if you return ```4.39353``` test won't accept that.

```
rate_service - method that adds a single rate to the wash station, and based on this single rate average_rating and count_of_ratings should be changed:
```

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list, income: float = 0) -> float:
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: callable) -> float:
+        calc_price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                      * self.average_rating / self.distance_from_city_center)
+
+        return round(calc_price, 1)
+
+    def wash_single_car(self, car: callable) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        avg_rating, count_ratings = self.average_rating, self.count_of_ratings
+        self.average_rating = round((avg_rating * count_ratings + rate)
+                                    / (count_ratings + 1), 1)
+        self.count_of_ratings += 1
--- a/app/main.py
+++ b/app/main.py
@@ -1,5 +1,7 @@
 class Car:
-    def __init__(self, comfort_class: int, clean_mark: int,
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
                  brand: str) -> None:
         self.comfort_class = comfort_class
         self.clean_mark = clean_mark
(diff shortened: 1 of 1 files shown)


oleksandra-shershen (approved):
LGTM!

app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
YuriiKindrat: it is better to place one/same number of args per line for better readability

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
YuriiKindrat: same is here

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+    def serve_cars(self, cars: list, income: float = 0) -> float:
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: callable) -> float:
+        calc_price = (car.comfort_class * (self.clean_power - car.clean_mark)
```
YuriiKindrat: use full names for vars, f.e. washing_price, or calculated_price or full_price

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list, income: float = 0) -> float:
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: callable) -> float:
```
YuriiKindrat: it seems to me, that car should be a Car class instance 
fix this everywhere

app/main.py
```diff
@@ -1,8 +1,38 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list, income: float = 0) -> float:
```
YuriiKindrat: write more descriptive type annotation for cars arg

YuriiKindrat (changes requested):

panicua:
-Changed args positioning in class ```CarWashStation``` methods
-Changed ```calc_price``` variable to ```calculated_price```
-Chanded type annotation to arguments "car" inside ```CarWashStation``` class. From ```car: callable```, to ```car: Car```.
-Chanded type annotation to argument "cars" inside ```CarWashStation``` class. From ```cars: list```, to ```cars: list[Car]```

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car], income: float = 0) -> float:
+        for car in cars:
+            if car.clean_mark < self.clean_power:
```
LiudmylaKulinchenko: car clean_mark should be less or equal to wash station clean_power

LiudmylaKulinchenko (changes requested):

app/main.py
panicua: I'll change it, but don't you think it doesn't have any logic? like why do we need to clean a car if it's ```clean_mark == clean_power```, it's gonna do nothing to cleanliness of the car. And also because of that our income not gonna change you can see the screenshot lower with calculations where ```clean_mark == clean_power```

NVIDIA_Share_iKUiap7D8l

panicua: NVIDIA_Share_50BM8RH4Gw
NVIDIA_Share_qE7uZ8ftG7
NVIDIA_Share_FsumFni5oy

LiudmylaKulinchenko (approved):
Good job 🔥

PR #1148, opened 2024-03-15
Solution

Gilgumish:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+            , 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
```
CharmNiovi: You check twice, the first time in line 28, and if it fits, the same check in the call to the wash_single_car method.

```
```

This will also work

CharmNiovi (changes requested):

YuriiKindrat (approved):
LGTM!

PR #1150, opened 2024-03-15
"Solution"

poplyasita:


viktoria-rybenchuk (changes requested):
All tests should passed, fix flake8

PR #1158, opened 2024-03-17
Solution#1

VitalyBashkiser:


app/main.py
```diff
@@ -1,8 +1,41 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
panicua: When Yurii checked my code he said that it's better to place one/same number of args per line for better readability. So I guess need to fix it.

app/main.py
```diff
@@ -1,8 +1,41 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
```
panicua, line 27: I don't know what version of Python you use but for me it works fine without ```from typing import List```, just ```cars: list[Car]```. My version of Python is 3.11, so if my approach doesn't work for you, don't need to fix it.

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        total_income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                wash_cost = self.calculate_washing_price(car)
+                total_income += wash_cost
+                self.wash_single_car(car)
+        return round(total_income, 1)
```
panicua: you don't need to ```round``` return here because you already rounded it in ```wash_cost = self.calculate_washing_price(car)``` method.

panicua (changes requested):

app/main.py
VitalyBashkiser: py 3.10: F821 undefined name 'List'

panicua (commented):
Looking good for me 👍

YuriiKindrat (approved):
LGTM!

PR #1163, opened 2024-03-18
class Car and class CarWashStation

semermaks:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: int,
```
Gilgumish: The condition says that the rating is not an integer, so it's better to use float here

Gilgumish (approved):

app/main.py
LiudmylaKulinchenko: I agree, type should be float

LiudmylaKulinchenko (changes requested):

YuriiKindrat (changes requested):
image
it seems to me, that u fixed not needed rating )

YuriiKindrat (approved):
GJ!

PR #1164, opened 2024-03-18
Success

Nyxan:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+
+    def serve_cars(self, cars: list) -> int:
+        total_income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(total_income, 1)
+
+    def calculate_washing_price(self, car: str) -> int:
```
y-kondrashova: These annotations do not match, you can easily check it using this `print(CarWashStation.calculate_washing_price.__annotations__)`

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        washing_price = (car.comfort_class
+                         * (self.clean_power - car.clean_mark)
+                         * (self.average_rating
+                            / self.distance_from_city_center))
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: str) -> bool:
```
y-kondrashova: Here with annotations the same

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> int:
```
y-kondrashova: This function does not return anything, check annotation here as well

y-kondrashova (changes requested):

app/main.py
Nyxan: I change annotations, but i think no need in return this function

app/main.py
Nyxan: No need in return this function too

y-kondrashova: I did not meant that. There was a point just with annotations. Everything else look good to me 😊

y-kondrashova (approved):
Good Job 😁

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
YuriiKindrat, line 19: u can use more descriptive type annotation here
`list[Car]`

YuriiKindrat (approved):

PR #1179, opened 2024-03-19
solution

NAZ-KOM:


app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change to the `CarWashStation` class as well

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
     pass
```
Dimosphen1: Delete `pass`, there is no need for it

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #1185, opened 2024-03-20
Solution: Car wash staition

AndrewChugreev:


app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change here and in the `CarWashStation` class as well

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #1190, opened 2024-03-27
solution

dvkim28:


dvkim28:
pls check

app/main.py
```diff
@@ -1,8 +1,46 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
+            -> None:
```
vsmutok: bad style code 

don't use `\`

vsmutok (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

app/main.py
EdAlekseiev: Still not fixed

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.clean_mark: int = clean_mark
+        self.brand: str = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) \
+            -> None:
```
EdAlekseiev: don't use `\`

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
v1

app/main.py
EdAlekseiev: Still not fixed
```suggestion
    def __init__(self, comfort_class: int, clean_mark: int, brand: str) -> None:
```

EdAlekseiev (changes requested):

app/main.py
dvkim28: @EdAlekseiev I have deleted "/" whats wrong now?

vsmutok (approved):
Looks Good!

PR #1197, opened 2024-03-30
The solution of the task

Ja0ni:


app/main.py
```diff
@@ -1,8 +1,53 @@
...
+    def serve_cars(self, cars: list) -> float:
+        price = sum(
+            self.calculate_washing_price(car)
+            for car in cars if car.clean_mark < self.clean_power
+        )
+
+        [
+            self.wash_single_car(car)
+            for car in cars if car.clean_mark < self.clean_power
+        ]
```
ArsenPidhoretskyi: No need to iterate two times, better to use forloop in this case

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change to `CarWashStation` class as well

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Dimosphen1: Indicate that you expect to obtain a list of `Car` class instances for `cars` in the type annotation

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1203, opened 2024-04-03
Solution

danilsiv:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            income = (car.comfort_class * (self.clean_power - car.clean_mark)
+                      * self.average_rating / self.distance_from_city_center)
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> None:
```
ArsenPidhoretskyi, line 32: Due to task requirements:

> wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark

If customer have car with clean_mark = 10, and our WashStation have clean_power = 5
If we do:
wash_station.wash_single_car(car)
As a result we will have car.clean_mark == 5
It is wrong, it should not change

ArsenPidhoretskyi (changes requested):
Well done!

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list[Car]) -> float:
+        income = 0.0
+        for car in cars_list:
+            if car.clean_mark < self.clean_power:
+                income += (car.comfort_class
+                           * (self.clean_power - car.clean_mark)
+                           * self.average_rating
+                           / self.distance_from_city_center)
```
ArsenPidhoretskyi: Reuse calculate_washing_price function

app/main.py
```diff
@@ -1,8 +1,44 @@
...
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list[Car]) -> float:
```
ArsenPidhoretskyi:
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```
Rename to cars

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1220, opened 2024-04-16
Solution

mykola-tarasenko:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
EdAlekseiev: You should specify inner type of list
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        income = []
+
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                income.append(self.calculate_washing_price(car))
+                self.wash_single_car(car)
+
+        return round(sum(income), 1)
```
EdAlekseiev: Why is income list? You can calculate income inside `for` loop.
```suggestion
        income = 0

        for car in cars:
            if self.clean_power > car.clean_mark:
                income += self.calculate_washing_price(car)
                self.wash_single_car(car)

        return round(income, 1)
```

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> None:
+        self.average_rating = round(sum(
+            [self.average_rating] * self.count_of_ratings + [rating])
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
EdAlekseiev: Why do you use `sum` here? You should use simple mathematical operations
```suggestion
        self.average_rating = round(
            (self.average_rating * self.count_of_ratings + rating)
            / (self.count_of_ratings + 1), 1)
        self.count_of_ratings += 1
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1229, opened 2024-04-25
add solution

midetc:


app/main.py
```diff
@@ -1,8 +1,74 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        washing_price_cars = 0
+        cars_can_wash = [car for car in cars if
+                         car.clean_mark < self.clean_power]
+        if cars_can_wash:
```
ArsenPidhoretskyi: This condition is redundant

app/main.py
```diff
@@ -1,8 +1,74 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        washing_price_single_car = (
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * (self.average_rating / self.distance_from_city_center)
+        )
+        return round(washing_price_single_car, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
ArsenPidhoretskyi: wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark

Condition is skipped

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: add condition here as well

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #1239, opened 2024-05-04
solution

necobra:


DevSpaciX (approved):

app/main.py
```diff
@@ -1,8 +1,39 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
```
DevSpaciX, line 2: For future please, don't use `* \` as a line break , you can add W504 flake8 rule to ignored ones or create a few temporal variables

PR #1241, opened 2024-05-07
'Solution'

Anon920:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                 distance_from_city_center: float,
+                 clean_power: float,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
EdAlekseiev: You should specify inner type of list
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

EdAlekseiev (changes requested):

app/main.py
Anon920: I did it, but I don't quite understand what it's for

EdAlekseiev: Specifying the inner type of a list in type hints makes your code more readable and easier for other developers to understand.

EdAlekseiev (approved):
Good job

PR #1246, opened 2024-05-10
Solution

BohdanKuzik:


app/main.py
```diff
@@ -1,8 +1,66 @@
...
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = (total_rating + rating) / self.count_of_ratings
+        self.average_rating = round(self.average_rating, 1)
+
+
+bmw = Car(3, 3, "BMW")
```
DevSpaciX: I see that you tested your code, but please add ONLY the solution to the task in the PR
Remove lines from 52 to 66
And also why did you deleted so many files ?) You should have only worked in the main file

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job ! But next time do not delete ANY files ))

PR #1249, opened 2024-05-13
py-car-wash-station v1

BornToLivee:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                wash_cost = self.calculate_washing_price(car)
+                income += wash_cost
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, cars: Car) -> float:
+        price = (cars.comfort_class * (self.clean_power - cars.clean_mark)
```
DevSpaciX: If you can do it in single line , always try to do it :)

DevSpaciX (changes requested):

app/main.py
ruslan-sketch: Привіт! То в мене помилка в формулі?

DevSpaciX: No-no ) All great , you did well !

DevSpaciX (approved):
Good job !

PR #1251, opened 2024-05-13
added solution to car wash station task

AlexRozv:
added implementation of Car and CarWashStation classes
implemented all methods inside CarWashStation

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
LiudmylaKulinchenko: better to move each parameter to a new line, including self

app/main.py
```diff
@@ -1,8 +1,39 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🔥

PR #1253, opened 2024-05-14
Solution

juliazelena:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return (car.comfort_class * (self.clean_power - car.clean_mark)
+                * self.average_rating / self.distance_from_city_center)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.validate_car(car):
```
pavlopro: You use 'wash_single_car' method only once. There is no need for this condition ('validate_car') here. You already made this check above

pavlopro (changes requested):
Several changes requested

vasylhnatiuk (approved):

PR #1258, opened 2024-05-16
Solution

SuskyiVolodymyr:


app/main.py
```diff
@@ -1,8 +1,55 @@
...
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
+        income = 0
+        if cars:
```
DevSpaciX: Here it is not necessary to check if cars exists, if there are empty list, then you simply will not be able to iterate through the loop and will immediately exit it , with a income value = 0

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+            for car in cars:
+                if car.clean_mark < self.clean_power:
+                    income += self.calculate_washing_price(car)
+                    self.wash_single_car(car)
+            return income
+        return 0
```
DevSpaciX: Just return income value and keep in mind comment above

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car:
```
DevSpaciX: Same , you just can return the rouned price here , without chekong if car exists

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+            price = ((car.comfort_class
+                      * (self.clean_power - car.clean_mark)
+                      * self.average_rating)
+                     / self.distance_from_city_center)
+            return round(price, 1)
+        return 0
+
+    def wash_single_car(self, car: Car) -> None:
+        if car:
```
DevSpaciX: Instead of this check , it should be - >
image

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice !

PR #1266, opened 2024-05-19
implemeth the solution

jpdreamthug:
All tests and flake8 passed
image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -5,6 +5,7 @@ class Car:
     def __init__(self, comfort_class: int,
                  clean_mark: int,
                  brand: str) -> None:
+
         self.comfort_class = comfort_class
         self.clean_mark = clean_mark
         self.brand = brand
@@ -12,10 +13,11 @@ class Car:
 
 class CarWashStation:
     def __init__(self,
-                 distance_from_city_center: int,
+                 distance_from_city_center: float,
                  clean_power: int,
                  average_rating: float,
                  count_of_ratings: int) -> None:
+
         self.distance_from_city_center = distance_from_city_center
         self.clean_power = clean_power
         self.average_rating = average_rating
@@ -23,26 +25,36 @@ class CarWashStation:
 
     def serve_cars(self, cars_list: list) -> Union[int, float]:
         result = 0
+
         for car in cars_list:
-            wash_cost = self.wash_single_car(car)
-            if wash_cost is not None:
-                result += wash_cost
-        return round(result, 1)
+
+            if self.clean_power > car.clean_mark:
+                wash_cost = self.calculate_washing_price(car)
+
+                if wash_cost is not None:
+                    result += wash_cost
+                    self.wash_single_car(car)
+
+        return result
 
     def calculate_washing_price(self, car: Car) -> Union[int, float]:
-        wash_cost = (car.comfort_class * (self.clean_power - car.clean_mark)
-                     * self.average_rating / self.distance_from_city_center)
-        return wash_cost
+        return round(
+            (car.comfort_class
+             * (self.clean_power
+                - car.clean_mark)
+             * self.average_rating
+             / self.distance_from_city_center), 1)
 
-    def wash_single_car(self, car: Car) -> Union[int, float]:
+    def wash_single_car(self, car: Car) -> None:
+
         if self.clean_power > car.clean_mark:
-            wash_cost = self.calculate_washing_price(car)
             car.clean_mark = self.clean_power
-            return wash_cost
-        return 0
 
-    def rate_service(self, number: int) -> None:
+    def rate_service(self, rate: int) -> None:
         self.count_of_ratings += 1
-        self.average_rating = round((self.average_rating
-                                    * (self.count_of_ratings - 1) + number)
-                                    / self.count_of_ratings, 1)
+        self.average_rating = round(
+            (self.average_rating
+             * (self.count_of_ratings
+                - 1)
+             + rate)
+            / self.count_of_ratings, 1)
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,3 @@
-from typing import Union
-
-
 class Car:
     def __init__(self, comfort_class: int,
                  clean_mark: int,
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,48 @@
+from typing import Union
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> Union[int, float]:
+        result = []
+        for car in cars_list:
+            wash_cost = self.wash_single_car(car)
+            if wash_cost is not None:
+                result.append(wash_cost)
+        return round(sum(result), 1)
+
+    def calculate_washing_price(self, car: Car) -> Union[int, float]:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> Union[int, float]:
+        result = []
```
DevSpaciX: You do not need to create list with numbers and then call sum() , you can create variable with start value 0 and on every iteration add wash cost to it

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> Union[int, float]:
+        result = 0
+        for car in cars_list:
+            wash_cost = self.wash_single_car(car)
+            if wash_cost is not None:
+                result += wash_cost
+        return round(result, 1)
```

According to the task description. on the screenshot attached, `serve_cars` in your implementation can perform required actions. However, since the task asks to return total cost ( even rounded ) it would be better to call a method for price calculation right from here.

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def calculate_washing_price(self, car: Car) -> Union[int, float]:
+        wash_cost = (car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center)
```
wQuelS, line 36: can you just make smaller chunks of your formula to avoid such a mess when trying to perform calculations in one action!?

Or at least make unified formatting for your calculations across the whole code base

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        return wash_cost
+
+    def wash_single_car(self, car: Car) -> Union[int, float]:
+        if self.clean_power > car.clean_mark:
+            wash_cost = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return wash_cost
+        return 0
```
wQuelS: by implementing a "calculation" call within your `serve_cars()` method you can get rid of two return statements. Also, `return 0` looks kinda weird. 
I assume that this method should wash a car and that's it

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def rate_service(self, number: int) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round((self.average_rating
+                                    * (self.count_of_ratings - 1) + number)
+                                    / self.count_of_ratings, 1)
```
wQuelS: as I mentioned above, the formatting style of your formulas differs. 
Splitting it into smaller and more readable chunks is also a good idea. It is up to you

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+            if wash_cost is not None:
+                result += wash_cost
+        return round(result, 1)
+
+    def calculate_washing_price(self, car: Car) -> Union[int, float]:
+        wash_cost = (car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center)
+        return wash_cost
+
+    def wash_single_car(self, car: Car) -> Union[int, float]:
```
wQuelS: seems like this method is not supposed to return anything

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        if self.clean_power > car.clean_mark:
+            wash_cost = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return wash_cost
+        return 0
+
+    def rate_service(self, number: int) -> None:
```
wQuelS: A `number` seems to be not very good for a variable name here. "rate", "rating", and "mark" -> those options look better, I guess

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
```
are you sure about the type hint here?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,60 @@
...
+        return round(
+            (car.comfort_class
+             * (self.clean_power
+                - car.clean_mark)
+             * self.average_rating
+             / self.distance_from_city_center), 1)
+
+    def wash_single_car(self, car: Car) -> None:
+
+        if self.clean_power > car.clean_mark:
```
DevSpaciX, line 40: You already have this check on 31 line , so you can remove it , is not it ?)

app/main.py
```diff
@@ -1,8 +1,60 @@
...
+
+    def serve_cars(self, cars_list: list) -> Union[int, float]:
+        result = 0
+
+        for car in cars_list:
+
+            if self.clean_power > car.clean_mark:
+                wash_cost = self.calculate_washing_price(car)
+
+                if wash_cost is not None:
```
DevSpaciX: Also you do not really need this check

DevSpaciX (changes requested):
Almost !))

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> Union[int, float]:
+        result = 0
+
+        cars_to_wash = [
```
DevSpaciX: Look now the complexity of your algorithm has increased due to the use of 2 loops, leave this simple check in one loop and that’s it

DevSpaciX (changes requested):
We don't need extra loops : )

app/main.py
jpdreamthug: What will happen if we use ```wash_single_car``` without ```serve_cars```?

jpdreamthug: I my opinion we need two checks in both methods

wQuelS: > What will happen if we use `wash_single_car` without `serve_cars`?

good point

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+
+        for car in cars_list:
+            if self.clean_power > car.clean_mark:
+                wash_cost = self.calculate_washing_price(car)
+                result += wash_cost
+                self.wash_single_car(car)
+
+        return result
+
+    def calculate_washing_price(self, car: Car) -> Union[int, float]:
```
A `distance from center` param is actually a part of the formula and we know that this variable's type is `float`.

Are u sure about your return type annotation?

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        return round(
+            (car.comfort_class
+             * (self.clean_power
+                - car.clean_mark)
+             * self.average_rating
+             / self.distance_from_city_center), 1)
```
wQuelS: To me it is a bit strange:  
- in the `serve_cars` method you have an additional variable `wash_cost` to add it lately to the `result`. 
- In methods with big formulas you've decided to perform everything in a single "turn".

In Python readability is very important. Your formula looks messy and it is not easy to understand what is going on there. I'd suggest splitting it into smaller chunks with additional variables. 

image

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
wQuelS: yeah, I see your point here: https://github.com/mate-academy/py-car-wash-station/pull/1266#discussion_r1607793594

It seems pretty reasonable for me, so you can also add that check to the `wash_single_car` method so it can be used separately.

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+    def serve_cars(self, cars_list: list) -> Union[int, float]:
+        result = 0
+
+        for car in cars_list:
+            if self.clean_power > car.clean_mark:
+                wash_cost = self.calculate_washing_price(car)
+                result += wash_cost
+                self.wash_single_car(car)
+
+        return result
```
wQuelS: You may also consider renaming this variable, since `result` is way too general name and it does not describe its goal

wQuelS (changes requested):

DevSpaciX (approved):
Good job !

PR #1272, opened 2024-05-20
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
```
DevSpaciX: What if you delete this chek ? Because You already check `if car.clean_mark < self.clean_power:` on line 23

DevSpaciX (changes requested):
Just a littlle optimisation requested

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+            car.clean_mark = self.clean_power
+        return self.calculate_washing_price(car)
```
DevSpaciX: Also you already calculating washing price on line 24 , so you do not need call it here

app/main.py
```diff
@@ -1,8 +1,40 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
```
wQuelS: what if input data consists of cars with `clean_mark` > `clean_power`.
Your `serve_cars()` method returns `0` in such case, which is not `float`

BileichukIvan:
Like this?

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = car.comfort_class * (self.clean_power - car.clean_mark) * \
+            (self.average_rating / self.distance_from_city_center)
```
wQuelS: please, avoid those line breaks.  You can wrap your code block with `()` to get rid of those  " \ ".

Also, readability in Python is pretty important. Such a messy formula can be chunked into few smaller parts.

wQuelS (changes requested):

app/main.py
BileichukIvan: so it can be solved by -> float | int??

app/main.py
BileichukIvan: So method should look like this?
```    
       comfort_class = car.comfort_class
       cleanliness_val = self.clean_power - car.clean_mark
       rating_val = self.average_rating / self.distance_from_city_center
       price = comfort_class * cleanliness_val * rating_val
       return round(price, 1)
```

BileichukIvan: Also should i change rate_service() as well??

BileichukIvan: So it also should look like this:
```
        total_ratings = self.count_of_ratings + 1
        average_rating = (((self.average_rating * self.count_of_ratings)
                           + rating) / total_ratings)
        self.average_rating = round(average_rating, 1)
        self.count_of_ratings = total_ratings
```
(i was forced to make a line break because of the length of a line)

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        rating_val = self.average_rating / self.distance_from_city_center
+        price = comfort_class * cleanliness_val * rating_val
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        total_ratings = self.count_of_ratings + 1
+        average_rating = (((self.average_rating * self.count_of_ratings)
```
DevSpaciX, line 40: Imho this way it is more readble 
```
average_rating = (
        (self.average_rating * self.count_of_ratings) + rating
    ) / total_ratings
```

DevSpaciX (approved):
All good but pay attention to the comment : )

app/main.py
wQuelS: > (i was forced to make a line break because of the length of a line)

as I've written in a previous comment, you can wrap the needed code block with parentheses `()` in such way:
```
price = (
    comfort_class 
    * (clean_power - clean_mark) 
    * (average_rating / distance_from_city_center)
)
```
So, the outer parentheses `()` in the example above have only one goal: formatting routine simplification which also includes possibility to avoid those line breaks " \\ "

PR #1274, opened 2024-05-20
car-wash-station solution

DimaYankovskiy:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
wQuelS: what if input data contains only cars with a `clean_mark` that is higher than the `clean_power` of the station?
Calling the `server_cars` method would return `0` in such case, which is `int`, not `float`.

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power
+                        - car.clean_mark)
+                     * (self.average_rating
+                        / self.distance_from_city_center), 1)
```
wQuelS: In cases with such huge formulas it is usually better to split them into smaller and more readable chunks

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        self.average_rating = round(
+            (self.average_rating
+             * self.count_of_ratings + new_rating)
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
wQuelS: same about splitting the formula into smaller parts.

Also, if you decide that formula is readable as it is and no splitting is needed - please, check how those formulas are formatted across your code and make a single variant

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        comfort_class = car.comfort_class
+        clean_power_difference = self.clean_power - car.clean_mark
+        average_rating = self.average_rating
+        distance_from_city_center = self.distance_from_city_center
```
wQuelS: only `clean_power_difference = self.clean_power - car.clean_mark` is reasonable. In other cases, you've just "removed"  `self.` from attributes. 

Having that  `clean_power_difference`, you can, for example, add variable for division part of the formula:
`<your_name_for_division> = self.average_rating / self.distance_from_city_center`.
The last action you would need is to multiplicate everything ( `car.comfort_class` and two new variables ) and round the result

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        total_rating_sum = self.average_rating * self.count_of_ratings
+        new_total_rating_sum = total_rating_sum + new_rating
+        new_count_of_ratings = self.count_of_ratings + 1
+        new_average_rating = new_total_rating_sum / new_count_of_ratings
+
+        self.average_rating = round(new_average_rating, 1)
+        self.count_of_ratings += 1
```
wQuelS: This formula is chunked much better. But, why do need that `new_count_of_ratings = self.count_of_ratings + 1`

You can simply increase your rate quantity attribute and use it to get the avg. rating

wQuelS (changes requested):

DevSpaciX (approved):
Good job !)

PR #1278, opened 2024-05-21
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        income = 0.0
```
DevSpaciX: Just let it be 0

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        cost = (
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+        )
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
DevSpaciX: This function already calling if - > `if car.clean_mark < self.clean_power:` Because you check it on 32 line

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice !

PR #1280, opened 2024-05-21
'Solution

Irina17191:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        total_income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income = self.calculate_washing_price(car)
+                # income = (car.comfort_class
```
DevSpaciX: Don't forget to clean up commented code, of course, if it doesn't make any sense to other developers : )

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> None:
+        self.average_rating = round((self.average_rating
```
DevSpaciX, line 40: To make the code more readable, break the counting logic into variables and operate with them : )

DevSpaciX (changes requested):

DevSpaciX (approved):
Great !

PR #1282, opened 2024-05-21
Done HomeTask!

TkachukTara:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, dirty_cars: list[Car]) -> float:
```
DevSpaciX, line 17: Could this function, in theory, return `int` as well?

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        income = (car.comfort_class
+                  * (self.clean_power - car.clean_mark)
+                  * self.average_rating
+                  / self.distance_from_city_center)
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        price = self.calculate_washing_price(car)
```
DevSpaciX: Pay attention to the name of the function, it does not involve any calculation here
image

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        car.clean_mark = self.clean_power
+        return price
+
+    def rate_service(self, rate: float) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round(((self.average_rating
```
DevSpaciX, line 39: To make the code more readable, break the counting logic into variables and operate with them : )
( Same on line 25 )

DevSpaciX (changes requested):

app/main.py
DevSpaciX: You can write both of types  int and float ->` int | float`

app/main.py
DevSpaciX: Did you saw my comment about `breaking the counting logic into variables` ?

DevSpaciX (changes requested):
Did you saw my comment about breaking the counting logic into variables ?
Also pay attention to all comments pls : )

PR #1283, opened 2024-05-21
'Solution'

Oleh323:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                total += self.wash_single_car(car)
+        return round(total, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        income = (car.comfort_class * (self.clean_power - car.clean_mark)
+                  * self.average_rating / self.distance_from_city_center)
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        price = self.calculate_washing_price(car)
```
DevSpaciX: Pay attention to the name of the function, it does not involve any calculation here
image

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
DevSpaciX, line 24: Could this function, in theory, return `int` as well?

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        return round(income, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        price = self.calculate_washing_price(car)
+        car.clean_mark = self.clean_power
+        return price
+
+    def rate_service(self, rate: float) -> None:
+        self.count_of_ratings += 1
+        self.average_rating = round(((self.average_rating
```
DevSpaciX: To make the code more readable, break the counting logic into variables and operate with them : )

DevSpaciX (changes requested):

DevSpaciX (changes requested):
Did you change the code ?

Oleh323:
> Ви змінили код?

Так, але я спочатку нажав на кнопку на фото,
а вже потім закомітив і запушив.
pain

app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str)\
+            -> None:
```
wQuelS: please, avoid those line breaks

app/main.py
```diff
@@ -1,8 +1,43 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
```
do you remember this hint? Fix your formatting style where needed

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
```
wQuelS: not a good idea to perform rounding here. Can you guess why?

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
wQuelS, line 24: https://github.com/mate-academy/py-car-wash-station/pull/1283/files#r1609304246
You've marked the above comment as "resolved" without fixes. If you think there is no need to change your code because the answer to the question in the comment above is "no, never" or smth like that, - just add an answer why do you think so.

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(total, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        income = (car.comfort_class * (self.clean_power - car.clean_mark)
+                  * self.average_rating / self.distance_from_city_center)
+        return round(income, 1)
```
wQuelS: Aren't you "rounding" the same thing twice here?

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+    def wash_single_car(self, car: Car) -> float:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        new_total_rating = total_rating + rate
+        new_middle_rating = new_total_rating / (self.count_of_ratings + 1)
+        self.average_rating = round(new_middle_rating, 1)
+        self.count_of_ratings += 1
```
wQuelS: Line 41: you are using `self.count_of_ratings + 1` without any changes to the `count_of_ratings` attribute itself

Line 43: you are actually increasing that `count_of_ratings` attribute

What is your reason to store increased value only at the end of the whole logic?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
```
DevSpaciX, line 13: Did you see the comment about good practice of formatting ?)

DevSpaciX (approved):
Next time pay attention to all comments , That problems are not critical witch I left in comments but do not repeat them again pls : )

DevSpaciX (approved):

PR #1284, opened 2024-05-21
Solution

b3v3kt0r:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
```
DevSpaciX: Is this function always returning `int` ?) If you call `round(total_income, 1)` think what type of data it can be also

DevSpaciX (changes requested):
Almost perfect !

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        total_income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income = self.calculate_washing_price(car)
+                total_income += income
+                self.wash_single_car(car)
+        return round(total_income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = ((car.comfort_class * (
```
DevSpaciX, line 31: Imho it would be better to break the counting logic into parts .
Keep that in mind next time : )

DevSpaciX (approved):
All good , but pay attention to the comment

PR #1285, opened 2024-05-21
Solution

ghost:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round((self.average_rating
```
DevSpaciX: To make the code more readable, break the counting logic into variables and operate with them : )
( Same on line 29 )

DevSpaciX (changes requested):

app/main.py
ghost: in theory yes, but we do have round(num, 1) in return for more accurate value

app/main.py
ghost: ok sure

app/main.py
DevSpaciX: I mean we can retrun both of types int and float -> `int | float `

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
+                if price is not None:
+                    income += price
+                    self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark >= self.clean_power:
```
DevSpaciX: You do not need this check actually here , beacuse you already have that check in `serve_cars func`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int | float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
+                if price is not None:
```
wQuelS: do you really need this check? What possible case scenario can lead to a `False` here?

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                    income += price
+                    self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
```
wQuelS, line 34: Aren't you rounding the same thing twice here?

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        total_rating = (self.average_rating
+                        * self.count_of_ratings)
+        total_rating = total_rating + rate
+        self.average_rating = round(total_rating
+                                    / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings = self.count_of_ratings + 1
```
wQuelS: Why do you have `self.count_of_ratings + 1` in your formula?  You **_don't store_** this value to be able to use `self.count_of_ratings`, but you actually **_do store_** it on the very next line. What is the point of doing so?

wQuelS (changes requested):

app/main.py
ghost: yes, i changed this lines so now it make more sense

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
wQuelS: Why have you changed it back to just a `float`?

Imagine such case:  the `clean_mark` attribute in each car from the list happens to be bigger than the `clean_power` of a washing station. According to the task description, this station can not wash such cars. So, it will not pass the `if car.clean_mark < self.clean_power:` condition and income will remain `0`. Even after `round(0, 1)`.
Guess what? `0` is not a `float`

wQuelS (changes requested):

DevSpaciX (approved):
Good job , but next time pay more attention to do not duplicate your logic ( round ) or smth

PR #1286, opened 2024-05-21
sol 1

IvanVahanov:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round(
+            (self.count_of_ratings * self.average_rating + rate)
+            / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
+
+# E       AttributeError: 'CarWashStation' object has no attribute 'serve_cars'
```
DevSpaciX: Delete this line , and it will be perfect )

DevSpaciX (changes requested):

app/main.py
IvanVahanov: just for shure, you mean last comment?

DevSpaciX: Yeah )

DevSpaciX (approved):

PR #1287, opened 2024-05-21
Solution

Andrii1000:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+           clean_power: int,
+           average_rating: float,
+           count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
DevSpaciX: What if input data consists of cars with `clean_mark > clean_power`.
Your `serve_cars()` method returns 0 in such case, which is not float

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        self.average_rating = round((self.average_rating
```
DevSpaciX: Maybe this way it is more readable ?

```
self.average_rating = round(
        (self.average_rating * self.count_of_ratings + rating) 
        / (self.count_of_ratings + 1), 
        1
    )
```

DevSpaciX (changes requested):
Just a little optimization requested : )

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+           self, distance_from_city_center: float,
```
DevSpaciX: Also pls fix all flake8 errors )

DevSpaciX (changes requested):

app/main.py
Andrii1000: I don`t understand what mean this comment. I need Any type to return?

Andrii1000: or i need total income 0.0?

DevSpaciX: > I don`t understand what mean this comment. I need Any type to return?

No , sorry , just think what type is `0` , because you can return it too )

Andrii1000: > > I don`t understand what mean this comment. I need Any type to return?
> 
> No , sorry , just think what type is `0` , because you can return it too )
0 type we can use as integer , bool etc. But in this moment the total_income variable is initialized as a float (total_income = 0), and the function returns the value of total_income after the loop (return round(total_income, 1)). the return value will be 0.0 since total_income is set to 0 and is not modified in the loop. And for this I think that type of 0 is float too. this my mind for that , is it true? or I don`t understand something?

Andrii1000: > > I don`t understand what mean this comment. I need Any type to return?
> 
> No , sorry , just think what type is `0` , because you can return it too )

0 type we can use as integer , bool etc. But in this moment the total_income variable is initialized as a float (total_income = 0), and the function returns the value of total_income after the loop (return round(total_income, 1)). the return value will be 0.0 since total_income is set to 0 and is not modified in the loop. And for this I think that type of 0 is float too. this my mind for that , is it true? or I don`t understand something?

DevSpaciX: I'm talking about a case when your check `if car.clean_mark < self.clean_power:` will be `False`
(And this is potentially possible)

So right return type here will be :

`float | int:`

DevSpaciX (changes requested):
Just a little thing about type hint : )

app/main.py
Andrii1000: oh, ok understand that , thanks.  now i add this type

Andrii1000: for moment when i write , idk that we can use two types

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: float) -> None:
+        self.average_rating = round(
```
DevSpaciX, line 41: To make the code more readable, break the counting logic into variables and operate with them : ) Thats for future

DevSpaciX (approved):
Great ! Pay attention for comment that i wrote please : )

app/main.py
DevSpaciX: Oh yeah , it can be 2 and more !)

PR #1288, opened 2024-05-21
Solution of task

YevaYe:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: (float, int),
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        # rom 1.0 to 10.0
```
DevSpaciX: Why did you leave this comments ?) Remove them please : )

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+        self.clean_power = clean_power
+
+        self.average_rating = average_rating
+        # from 1.0 to 5.0, rounded to 1 decimal
+
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,8 +1,39 @@
...
-    pass
+    def __init__(self, distance_from_city_center: (float, int),
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> (float, int):
```
DevSpaciX, line 18: You can simply write retrun type hint like this -> `float | int `

DevSpaciX (approved):
Good , but pay attention to the comment )

PR #1290, opened 2024-05-21
'Solution'

MaglorFeanorson:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.count_of_ratings += 1
+        total_rating = self.average_rating * (self.count_of_ratings - 1)
+        temp_rating = (total_rating + rating) / self.count_of_ratings
+        self.average_rating = round(temp_rating, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
DevSpaciX: Can you theoretically return in this function `int` type ?

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.count_of_ratings += 1
+        total_rating = self.average_rating * (self.count_of_ratings - 1)
+        temp_rating = (total_rating + rating) / self.count_of_ratings
+        self.average_rating = round(temp_rating, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list[Car]) -> int:
```
DevSpaciX: I meant that you can return both float and int, right?

DevSpaciX (changes requested):

app/main.py
MaglorFeanorson: teoretically yeap

DevSpaciX (approved):
Great !

PR #1295, opened 2024-05-24
Test solution

VasylRysan:


app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

vasylhnatiuk (changes requested):

app/main.py
VasylRysan: As i understand, this meas that input list must contain only instances of the class "Car". But it looks like i missed the lesson, where such annotation specification was shown? I have not seen and used this before.

vsmutok (approved):
Good Job!

PR #1305, opened 2024-06-03
created car wash station class and car class

Timur5050:


app/main.py
```diff
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change to the `CarWashStation` class as well

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                        * (self.clean_power - car_to_wash.clean_mark)
+                        * self.average_rating / self.distance_from_city_center)
+        return round(sum_for_wash, 1)
+
+    def wash_single_car(self, car_to_wash: Car) -> None:
+        if self.clean_power >= car_to_wash.clean_mark:
+            car_to_wash.clean_mark = self.clean_power
+
+    def rate_service(self, new_rate: int) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
```
pavlopro: No need to have this variable
Do `average_rating` calculation in one line

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        total_rating += new_rate
+        self.count_of_ratings += 1
+        self.average_rating = round(total_rating / self.count_of_ratings, 1)
+
+    def serve_cars(self, list_of_cars: list[Car]) -> int:
+        total_sum = 0
+        for car in list_of_cars:
+            if self.clean_power >= car.clean_mark:
```
pavlopro, line 48: The code check already exists in 'wash_single_car'. This is a code duplication. Please remove it from here.

pavlopro (changes requested):

Timur5050:
I know that this if statement should be in the wash_single_car function, not in the serve_cars function, but if I do this, I will have problems with calculate_washing_priceof the car wash, because the mark about cleaning the car will be changed before the price for its washing is calculated

viktoria-rybenchuk (approved):
GJ!

PR #1306, opened 2024-06-04
Solution

MishaCherpak:


vasylhnatiuk (changes requested):
fix tests

app/main.py
```diff
@@ -1,8 +1,44 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: int, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_car: list) -> float:
```
EdAlekseiev: Speciffy inner type of list
```suggestion
    def serve_cars(self, list_car: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        profit = 0
+        for car in list_car:
+            if car.clean_mark < self.clean_power:
+                profit += CarWashStation.calculate_washing_price(self, car)
```
EdAlekseiev: Why do you use CarWashStation? You should use `self`. Fix everywhere
```suggestion
                profit += self.calculate_washing_price(self, car)
```

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                CarWashStation.wash_single_car(self, car)
+
+        return round(profit, 1)
+
+    def calculate_washing_price(self, car: Any) -> float:
```
```suggestion
    def calculate_washing_price(self, car: Car) -> float:
```

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        result = round(car.comfort_class
+                       * (self.clean_power - car.clean_mark)
+                       * (self.average_rating
+                          / self.distance_from_city_center), 1)
+        return result
+
+    def wash_single_car(self, car: Any) -> None:
```
```suggestion
    def wash_single_car(self, car: Car) -> None:
```

EdAlekseiev (changes requested):

app/main.py
MishaCherpak: if I change to self then it doesn't see the argument car

app/main.py
```diff
@@ -1,8 +1,44 @@
+from typing import Any
```
EdAlekseiev: Delete unused import, please

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_car: list[Car]) -> float:
+        profit = 0
+        for car in list_car:
+            if car.clean_mark < self.clean_power:
+                profit += CarWashStation.calculate_washing_price(self, car)
```
```suggestion
                profit += self.calculate_washing_price(car)
```

EdAlekseiev (changes requested):

app/main.py
MishaCherpak: Now I'll know that I can call the function this way, thanks

vsmutok (approved):
Great!

PR #1313, opened 2024-06-10
New right Solution

Postpostman:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self,
+                     new_rating: float) -> None:
+        total_rating_sum = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        total_rating_sum += new_rating
+        self.average_rating = \
```
vasylhnatiuk: try to avoid \

vasylhnatiuk (changes requested):

app/main.py
Postpostman: > try to avoid \

I tried, but flake8 do not pass cause of "line is too long"

```suggestion
        self.average_rating = round(
            total_rating_sum / self.count_of_ratings, 1
        )
```

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: use like here

```
        self.average_rating = round(
            total_rating_sum / self.count_of_ratings, 1
        )
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1318, opened 2024-06-11
Перезалив car wash station

RToister:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
+        # Відстань до центру міста
```
vasylhnatiuk: don't use comments to describe your code, use docs if you need explanation and please write docs in English

vasylhnatiuk (changes requested):

app/main.py
RToister: Ok, i will change

PR #1322, opened 2024-06-12
+

AndrZhuk:


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            (car.comfort_class
+             * (self.clean_power - car.clean_mark)
+             * self.average_rating
+             / self.distance_from_city_center)
+            , 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
```
pavlopro: Please remove this condition, as this method is used only in one place where this check is already present.

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1328, opened 2024-06-14
implement class Car and class CarWashStation

HalynaRohatska:
implement class Car and class CarWashStation with methods: serve_cars, calculate_washing_price, wash_single_car and rate_service.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -27,7 +27,9 @@ class CarWashStation:
         if 1.0 <= distance_from_city_center <= 10.0:
             self.distance_from_city_center = distance_from_city_center
         else:
-            raise ValueError("distance_from_city_center should be between 1.0 and 10.0")
+            raise ValueError(
+                "distance_from_city_center should be between 1.0 and 10.0"
+            )
         self.clean_power = clean_power
         if 1.0 <= average_rating <= 5.0:
             self.average_rating = round(average_rating, 1)
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,50 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list) -> float:
+        income = 0
+        for car in list_of_cars:
+            if car.clean_mark < self.clean_power:
+                income += round(self.calculate_washing_price(car), 1)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        difference_cleaning = self.clean_power - car.clean_mark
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,50 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
```

MNDonut (changes requested):

app/main.py
HalynaRohatska: added validate

app/main.py
```diff
@@ -1,8 +1,64 @@
...
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        if 1 <= comfort_class <= 7:
+            self.comfort_class = comfort_class
+        else:
+            raise ValueError("comfort_class should be between 1 and 7")
```
oliasymak24: try replacing the duplicate code

app/main.py
```diff
@@ -1,8 +1,64 @@
...
+        if 1 <= clean_mark <= 10:
```
oliasymak24, line 12: try replacing the duplicate code

app/main.py
```diff
@@ -1,8 +1,64 @@
...
+            self.clean_mark = clean_mark
+        else:
+            raise ValueError("clean_mark should be between 1 and 10")
+        self.brand = brand
```
oliasymak24, line 30: try replacing the duplicate code

app/main.py
```diff
@@ -1,8 +1,64 @@
...
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        if 1.0 <= distance_from_city_center <= 10.0:
+            self.distance_from_city_center = distance_from_city_center
+        else:
+            raise ValueError(
+                "distance_from_city_center should be between 1.0 and 10.0"
+            )
+        self.clean_power = clean_power
```
oliasymak24, line 47: try replacing the duplicate code

app/main.py
```diff
@@ -1,8 +1,64 @@
...
+        if 1.0 <= average_rating <= 5.0:
+            self.average_rating = round(average_rating, 1)
+        else:
+            raise ValueError("average_rating should be between 1.0 and 5.0")
```
oliasymak24: try replacing the duplicate code

oliasymak24 (changes requested):
you correctly understood your mistake, but try to do it another way (left a comment)

app/main.py
HalynaRohatska: I feel like there is another way to test the arguments, but I don't know what

app/main.py
```diff
@@ -1,18 +1,19 @@
+def validate_func(value: float, min_value: float, max_value: float, name_argument: str) -> int | float:
+    if min_value <= value <= max_value:
+        return value
+    else:
+        raise ValueError(f"{name_argument} should be between {min_value} and {max_value}")
+
```
MNDonut: The last reject, I swear :)
Note that if statement has return, it means you don't need else statement because after return python will not run the code after if statement even if there isn't else statement ;)

MNDonut (changes requested):

app/main.py
HalynaRohatska: I removed it.

oliasymak24 (approved):
You correct understand all comments)

PR #1334, opened 2024-06-20
completed washing station task, with creating Car and CarWashStation …

MasakDirt:
…classes and methods for the last one

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -3,8 +3,13 @@ class Car:
                  clean_mark: int, brand: str) -> None:
         if 1 <= comfort_class <= 7:
             self.comfort_class = comfort_class
+        else:
+            self.comfort_class = 1
+
         if 1 <= clean_mark <= 10:
             self.clean_mark = clean_mark
+        else:
+            self.clean_mark = 1
 
         self.brand = brand
 
@@ -15,8 +20,13 @@ class CarWashStation:
                  count_of_ratings: int) -> None:
         if 1.0 <= distance_from_city_center <= 10.0:
             self.distance_from_city_center = distance_from_city_center
+        else:
+            self.distance_from_city_center = 1.0
+
         if 1.0 <= average_rating <= 5.0:
             self.average_rating = average_rating
+        else:
+            self.average_rating = 1.0
 
         self.clean_power = clean_power
         self.count_of_ratings = count_of_ratings
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```

MNDonut (changes requested):

app/main.py
MasakDirt: Is it okay to use `if` statements in constructors?

app/main.py
```diff
@@ -1,8 +1,52 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        if 1 <= comfort_class <= 7:
+            self.comfort_class = comfort_class
+        if 1 <= clean_mark <= 10:
+            self.clean_mark = clean_mark
```
MNDonut: what if I set 11 for each field? It will not set any attribute, thus class instance won't have them and python will raise an error that class attribute doesn't have these attributes

app/main.py
```diff
@@ -1,8 +1,52 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        if 1.0 <= distance_from_city_center <= 10.0:
+            self.distance_from_city_center = distance_from_city_center
+        if 1.0 <= average_rating <= 5.0:
+            self.average_rating = average_rating
```
MNDonut: what if I set 11 for each field? It will not set any attribute, thus class instance won't have them and python will raise an error that class attribute doesn't have these attributes

MNDonut (changes requested):

app/main.py
MNDonut: Yes, it's. You can do whatever you want in class methods :)

app/main.py
MNDonut: > Is it okay to use `if` statements in constructors?

Yes, it's. You can do whatever you want in class methods :)

MasakDirt: Okay, thanks

MNDonut (approved):

PR #1341, opened 2024-06-25
Solution

NazarSerdiuk1:


viktoria-rybenchuk (changes requested):
All tests should passed, fix flake8

PR #1356, opened 2024-07-02
Solution

andrewsiemakin:


app/main.py
```diff
@@ -1,8 +1,39 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
+            -> None:
```
vsmutok: better to use `()` then `\`

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
```
```suggestion
    def serve_cars(self, cars: list[Car]) -> float:
```

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+                total_income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(total_income, 1)
+
+    def rate_service(self, new_rating: int) -> None:
+        total_rating_sum = self.average_rating * self.count_of_ratings
+        total_rating_sum += new_rating
+        self.count_of_ratings += 1
+        self.average_rating = \
+            round(total_rating_sum / self.count_of_ratings, 1)
```
vsmutok: better to use `()` then `\`

vsmutok (changes requested):

app/main.py
andrewsiemakin: Can you explain? I'm trying to move the code to the next line, because when I check with pep8 I get an error about the line length being exceeded. The IDE itself substituted "/" during the transfer.

app/main.py
andrewsiemakin: Can you explain? I'm trying to move the code to the next line, because when I check with pep8 I get an error about the line length being exceeded. The IDE itself substituted "/" during the transfer.

app/main.py
Dimosphen1: It is a good practice to avoid line-breaking operator `\` when defining a method's signature. Prefer placing each argument including `self` on a new line so there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change here and in the `__init__` for `CarWashStation` class as well

app/main.py
Dimosphen1: Place `round(` on the same line with `=` , `total_rating_sum / self.count_of_ratings, 1` on the next line and `)` on the third line so that closing parenthesis would not have any arguments on the same line as well as an opening parenthesis. In this way you could avoid using line-breaking `\` operator

app/main.py
Dimosphen1: Apply the mentioned change

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1359, opened 2024-07-05
Implemented solution

dimak20:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
+            -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```

app/main.py
```diff
@@ -1,8 +1,38 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: int, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
+            -> None:
+        if 1 <= comfort_class <= 7:
+            self.comfort_class = comfort_class
+        if 1 <= clean_mark <= 10:
+            self.clean_mark = clean_mark
```
MNDonut: What if some user runs your program via terminal(let's imagine we can do it) and enters 100, how you program will behave? As you set these attributes only when these conditions are met, it means in our case the class instance won't have them and user will get AttributeError when other methods will be called. So you should handle this case

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        if 1.0 <= distance_from_city_center <= 10.0:
+            self.distance_from_city_center = distance_from_city_center
+        if 1 <= clean_power <= 10:
+            self.clean_power = clean_power
+        if 1.0 <= average_rating <= 5.0:
+            self.average_rating = average_rating
+        if count_of_ratings > 0:
+            self.count_of_ratings = count_of_ratings
```
MNDonut: What if some user runs your program via terminal(let's imagine we can do it) and enters 100, how you program will behave? As you set these attributes only when these conditions are met, it means in our case the class instance won't have them and user will get AttributeError when other methods will be called. So you should handle this case

MNDonut (changes requested):

dimak20:
Considered abovementioned case. I did not know whether I was allowed to raise ValueError immediately in __init__ method.  Hope this approach may be correct

app/main.py
```diff
@@ -1,8 +1,61 @@
...
+            raise ValueError(
+                "The value of `clean_power` must be in the range from "
+                "1 to 10"
+            )
+        self.clean_power = clean_power
+        if not (1.0 <= average_rating <= 5.0):
+            raise ValueError(
+                "The value of `average_rating` must be in the range from "
+                "1.0 to 5.0"
+            )
```
MNDonut: Yea, you can do whatever you want in init. It's always better to do validation before you create some object with incorrect data. Or you can create a separate method with some parameters and call it inside init method.
Do you see how your code became harder to read? Also note that all if statements have a common pattern and it means you can create some method in order to not copy paste almost the same code

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.average_rating = round(validate(average_rating, 1.0, 5.0), 1)
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def serve_cars(self, cars: list) -> float:
```
oliasymak24, line 25: Will be better to add type annotation to the list of carsю

oliasymak24 (approved):
You changed everything correct, now your code looks much more better, good job!
P.s. left small comment)

PR #1363, opened 2024-07-08
Solution

SerhiyDema:
attempt 01

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -19,10 +19,10 @@ class CarWashStation:
         self.average_rating = average_rating  # from 1.0 to 5.0
         self.count_of_ratings = count_of_ratings  # number of voting
 
-    """
-    returns income of CarWashStation for serving list of Car's
-    """
     def serve_cars(self, list_cars: list[Car]) -> float:
+        """
+        returns income of CarWashStation for serving list of Car's
+        """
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        self.distance_from_city_center = distance_from_city_center  # 1.0 to 10
+        self.clean_power = clean_power  # up to clean_mark
+        self.average_rating = average_rating  # from 1.0 to 5.0
+        self.count_of_ratings = count_of_ratings  # number of voting
+
+    """
+    returns income of CarWashStation for serving list of Car's
+    """
+    def serve_cars(self, list_cars: list[Car]) -> float:
+        return round(sum([self.wash_single_car(car) for car in list_cars]), 1)
```
EdAlekseiev: Docstring should be inside function. Fix everywhere.
```suggestion
    def serve_cars(self, list_cars: list[Car]) -> float:
        """
        returns income of CarWashStation for serving list of Car's
        """
        return round(sum([self.wash_single_car(car) for car in list_cars]), 1)
```

EdAlekseiev (changes requested):

SerhiyDema:
Fix docstring inside function everywhere.

EdAlekseiev (approved):

PR #1365, opened 2024-07-11
Solution

OleksiukStepan:
Add and implement:

1. class Car:
- def __init__()

2. class CarWashStation:
- def __init__()
- def serve_cars
- def calculate_washing_price
- def wash_single_car
- def rate_service

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -7,7 +7,7 @@ class Car:
     ) -> None:
         _check_value("comfort_class", comfort_class, 1, 7)
         _check_value("clean_mark", clean_mark, 1, 10)
-        
+
         self.comfort_class = comfort_class
         self.clean_mark = clean_mark
         self.brand = brand
@@ -20,7 +20,12 @@ class CarWashStation:
             average_rating: float,
             count_of_ratings: int
     ) -> None:
-        _check_value("distance_from_city_center", distance_from_city_center, 1.0, 10.0)
+        _check_value(
+            "distance_from_city_center",
+            distance_from_city_center,
+            1.0,
+            10.0
+        )
         _check_value("average_rating", average_rating, 1.0, 5.0)
 
         self.distance_from_city_center = distance_from_city_center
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,53 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut, line 12: Validate values that are mentioned in the requirements

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+    def __init__(
+            self, distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut, line 34: Validate values that are mentioned in the requirements

MNDonut (changes requested):

app/main.py
OleksiukStepan, line 1: add method _check_value that check value according to conditions and if value is not valid shows appropriate message

app/main.py
```diff
@@ -1,8 +1,76 @@
...
+def _check_value(
+        value_name: str,
+        value: int | float,
+        lower_limit: int | float,
+        upper_limit: int | float
+) -> None:
+    if not (lower_limit <= value <= upper_limit):
+        raise ValueError(
+            f"{value_name} must be between {lower_limit} and {upper_limit}"
+        )
```
MNDonut, line 76: In case two classes use the same logic, you can create a base class that has that method and inherit from this class for both classes

MNDonut (approved):

app/main.py
OleksiukStepan: I have to create base class or Mixin? What will be better?

PR #1379, opened 2024-07-17
solution

georgiy-nezbritskiy:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int, brand: str
+    ) -> "Car":
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
```
LiudmylaKulinchenko: You don't have return statement in this method, so your return annotation is wrong. Fix it, please!

LiudmylaKulinchenko: Same everywhere

LiudmylaKulinchenko (changes requested):

app/main.py
georgiy-nezbritskiy: good catch!
but what is  the point of adding everywhere annotation  and still being able to return any types?

LiudmylaKulinchenko: It helps to understand easier what function/method does and how to use it. Also pycharm actually highlights this, so at the same time it's helpfull for developing as well

LiudmylaKulinchenko (approved):
Good job 🌸

PR #1382, opened 2024-07-18
solution

BastovOleksandr:


app/main.py
```diff
@@ -1,8 +1,52 @@
...
+
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+            , 1)
```
LiudmylaKulinchenko: fix formatting here. Since you have a line break after the first parentheses, you should have it before the last one

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, new_review: float) -> None:
+        self.average_rating = round((
+            (self.average_rating * self.count_of_ratings + new_review)
+            / (self.count_of_ratings + 1)
+        ), 1)
```
LiudmylaKulinchenko: You can also have a line break after `round(` to make a similar formatting throughout the code

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+            , 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```

LiudmylaKulinchenko (changes requested):

Oleksl888 (approved):
Well done!

PR #1385, opened 2024-07-18
two classes were realized with their methods

Narberal90:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        if 1 <= comfort_class <= 7:
+            self.comfort_class = comfort_class
+
+        if 1 <= clean_mark <= 10:
+            self.clean_mark = clean_mark
```
MNDonut: what if value is incorrect? it means class will not have these attributes at all

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        if 1.0 <= distance_from_city_center <= 10.0:
+            self.distance_from_city_center = distance_from_city_center
+
+        self.clean_power = clean_power
+
+        if 1.0 <= average_rating <= 5.0:
+            self.average_rating = average_rating
+
```
MNDonut: what if value is incorrect? it means class will not have these attributes at all

app/main.py
```diff
@@ -1,8 +1,54 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
MNDonut: let's follow the first option from checklist where we move all parameters on the next lines

app/main.py
```diff
@@ -1,8 +1,54 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
MNDonut, line 16: let's follow the first option from checklist where we move all parameters on the next lines

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+
+        if 1.0 <= average_rating <= 5.0:
+            self.average_rating = average_rating
+
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
```
MNDonut: let's follow the first option from checklist where we move all parameters on the next lines

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        if self.count_of_ratings == 0:
+            self.count_of_ratings = 1
+            self.average_rating = rate
+        self.average_rating = round((self.average_rating
+                                     * self.count_of_ratings + rate)
+                                    / (self.count_of_ratings + 1), 1)
```
MNDonut: let's follow the first option from checklist where we move all parameters on the next lines

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        if self.count_of_ratings == 0:
+            self.count_of_ratings = 1
+            self.average_rating = rate
```
MNDonut: I suppose you handle this case because of deletion by zero, but why do you that if when you divide, you have (self.count_of_ratings + 1), thus it'll never be zero. So you can simplify this method :)

MNDonut (changes requested):

app/main.py
Narberal90: you mean in this case everything is good? like good example?

app/main.py
Narberal90: I thought we have to control for producing only objects with good parameters.

app/main.py
LiudmylaKulinchenko: Ihor suggested you to use first example form the checklist in Linter formatter task, since this type of formatting is more common
https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md

LiudmylaKulinchenko (approved):
Well done 🌸 Also answered your question

PR #1391, opened 2024-07-21
'Solution'

dartomOOv:


app/main.py
```diff
@@ -1,8 +1,58 @@
...
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut, line 10: In the requirements you have specific values, you should check if correct values are passed. If not - set some default

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
```
MNDonut, line 25: In the requirements you have specific values, you should check if correct values are passed. If not - set some default

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        overall = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                overall += self.wash_single_car(car)
```
MNDonut: you can try to make a list comprehension here

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+        price = self.calculate_washing_price(car)
+        car.clean_mark = self.clean_power
+        return price
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(
+            (
+                ((self.average_rating * float(self.count_of_ratings))
+                 + rate) / float(self.count_of_ratings + 1)
+            ), 1)
```
MNDonut: In order you make it easier to read, I'd recommend you to split this statement on a few variables and then calculate the result

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+    def serve_cars(self, cars: list) -> float:
+        overall = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                overall += self.wash_single_car(car)
+
+        return overall
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (float(car.comfort_class)
```
MNDonut: move float... on the next line

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+                 * float(self.clean_power - car.clean_mark)
+                 * self.average_rating
+                 / self.distance_from_city_center
+                 )
```
MNDonut: We recommend to use the first approach from checklist, so let's fix it

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+
+        return sum(
+            [self.wash_single_car(car)
+             for car in cars
+             if car.clean_mark < self.clean_power]
```
LiudmylaKulinchenko: no need to use [], since sum can work with generators

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+        return price
+
+    def rate_service(self, rate: float) -> None:
+        summ_rates = self.average_rating * float(self.count_of_ratings) + rate
+        new_avr_rate = summ_rates / float(self.count_of_ratings + 1)
+
+        self.average_rating = round(new_avr_rate, 1)
+        self.count_of_ratings += 1
+
+        return
```
LiudmylaKulinchenko: No need to use blank return here

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+            * self.average_rating
+            / self.distance_from_city_center
+        )
+
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        price = self.calculate_washing_price(car)
+        car.clean_mark = self.clean_power
+        return price
```

LiudmylaKulinchenko (changes requested):
Please, don't mark comments as resolved. It's important to see previous comments in order to make you re-review, no need to close them during studying

app/main.py
dartomOOv: Is'n it going to check if clean_power is greater than clean_mark on the line 33? Or should I move it to the wash_single_car func?

dartomOOv:
I didn't read the question carefully, sorry for my stupid question ;)

LiudmylaKulinchenko (approved):
Good job 🤠 And remember there's no stupid questions!

PR #1393, opened 2024-07-21
Solution

mnezbrytska:


app/main.py
```diff
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
MNDonut: we recommend to follow the first option from checklist where we move all parameters on the next lines, so let's do it :)

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
MNDonut: we recommend to follow the first option from checklist where we move all parameters on the next lines, so let's do it :)

app/main.py
```diff
@@ -1,8 +1,40 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut, line 11: note that these attributes can have only specific values (you can find them in the requirements). So let's check it and if it's incorrect value, then set some default

app/main.py
```diff
@@ -1,8 +1,40 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut, line 29: note that these attributes can have only specific values (you can find them in the requirements). So let's check it and if it's incorrect value, then set some default

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+            if single_car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(single_car)
+                self.wash_single_car(single_car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = ((car.comfort_class
+                 * (self.clean_power - car.clean_mark)
+                 * self.average_rating)
+                 / self.distance_from_city_center)
```
MNDonut: we recommend to follow the first option from checklist where we move all parameters on the next lines, so let's do it :)

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,52 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
Oleksl888: Please fix so that every parameter starts on a separate line:
```
def __init__(
    self, 
    comfort_class: int...
or
def __init__( self, 
              comfort_class: int...

File: app/main.py [hunk 7]
```diff
@@ -1,8 +1,52 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float,
```
Oleksl888: Same as above - please fix parameters to start on new line

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for single_car in cars:
+            if single_car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(single_car)
+                self.wash_single_car(single_car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = ((car.comfort_class
```
Oleksl888: Here you don't need to store price in a variable (since you don't reuse it)  - you can return straight away.

app/main.py
Oleksl888: Please set the default value for attribute in case if the value is incorrect. 
(Actually the correct implementation here would be using properties to validate attribute data but we will study it later).
For now replace raising an exception to setting a default value as per previous comment.

Oleksl888 (changes requested):

oliasymak24 (approved):
Good job!

PR #1396, opened 2024-07-22
Solution

mihavryliuk:


app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
```
Oleksl888: Please fix as per styling guidelines - each param must be on new line, also we don't use `\` as a line breaker

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+            -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
```
Oleksl888: Please fix as per styling best practices

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark >= self.clean_power:
```
Oleksl888: there is no requirement to check this here

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+            return 0.0
+        price = (
```
Oleksl888: no need to create an extra variable that you will not reuse - just do all calculations in return statement

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+        ) / self.distance_from_city_center
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
```
Oleksl888: you can be more specific about list contents data type

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                self.wash_single_car(car)
+                total_income += wash_price
+        return round(total_income, 1)
+
+    def rate_service(self, new_rating: float) -> None:
+        total_ratings = self.count_of_ratings + 1
+        new_average_rating = (
+            (self.average_rating * self.count_of_ratings) + new_rating
+        ) / total_ratings
+        self.average_rating = round(new_average_rating, 1)
```
Oleksl888: you can set new average rating in one statement without creating extra variables

Oleksl888 (changes requested):

app/main.py
mihavryliuk: what i dont need to check???

mihavryliuk: if?

Oleksl888: There is no requirement to check clean power in this method as per task definition

> calculate_washing_price - method, that calculates cost for a single car wash, cost is calculated as: car's comfort class * difference between wash station's clean power and car's clean mark * car wash station rating / car wash station distance to the center of the city, returns number rounded to 1 decimal;

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        total_income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                wash_price = self.calculate_washing_price(car)
+                self.wash_single_car(car)
+                total_income += wash_price
```
Oleksl888, line 46: Here you could just do so you don't have to create an extra variable and line of code:
```

File: app/main.py [hunk 8]
```diff
@@ -1,8 +1,54 @@
...
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
```
Oleksl888, line 27: Here you could do like so so you don't use extra parenthesis:
```
return round(
            car.comfort_class
            * self.average_rating
            / self.distance_from_city_center, 
           1
        )

Oleksl888 (approved):
Well done!

PR #1397, opened 2024-07-22
Solution

panichqa:


app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
MNDonut: Let's always use the first option from checklist where we move all parameters on the next lines

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+        self, distance_from_city_center: float, clean_power: int,
+        average_rating: float, count_of_ratings: int
+    ) -> None:
```
MNDonut: Let's always use the first option from checklist where we move all parameters on the next lines

app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: In the requirements these attributes have a range of values they can accept. Let's check values and if they are incorrect, set some default value and is correct

app/main.py
```diff
@@ -1,8 +1,42 @@
...
-    # write your code here
-    pass
+    def __init__(
+        self, distance_from_city_center: float, clean_power: int,
+        average_rating: float, count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: In the requirements these attributes have a range of values they can accept. Let's check values and if they are incorrect, set some default value and is correct

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round((total_rating + rate)
+                                    / self.count_of_ratings, 1)
```
MNDonut: Let's follow the first approach from the checklist
```
round(
    something
, 1)
```

MNDonut: Also we don't use \ as a line breaker. It's better to use parenthesis

MNDonut (changes requested):

.flake8
Oleksl888, line 1: please rollback changes from this file

app/main.py
```diff
@@ -1,8 +1,79 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+        self, comfort_class: int, clean_mark: int, brand: str
```
Oleksl888: When you have multiple params it is a best practice to have each parameter on new line

app/main.py
Oleksl888: This hasn't been changed

app/main.py
```diff
@@ -1,8 +1,79 @@
...
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = (
```
Oleksl888: No need to store result of calculation in variable since you are not going to reuse it.
```
return round(
            car.comfort_class
            ...

File: app/main.py [hunk 8]
```diff
@@ -1,8 +1,79 @@
...
+            * self.average_rating / self.distance_from_city_center
+        )
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
```
Oleksl888: Can you rewrite this formula so you don't have to use total_rating variable ?

Oleksl888 (changes requested):

.flake8
panichqa: my flake8 doesnt work without this change

Oleksl888, line 1: You cannot make changes to files other then specified in task requirements. 
It seems you need this change because your local environment has different name for virtual env, but you don't have to push these changes to remote

Oleksl888 (changes requested):
Revert the changes to `.flake8` and make new push to remote without those changes

app/main.py
```diff
@@ -1,8 +1,83 @@
...
+
+    def validate_comfort_class(self, comfort_class: int) -> int:
+        if 1 <= comfort_class <= 7:
+            return comfort_class
+        return 1
+
+    def validate_clean_mark(self, clean_mark: int) -> int:
+        if 1 <= clean_mark <= 10:
+            return clean_mark
+        return 1
```
MNDonut: You can see that these methods look similar, so you can create a common function and use it for each attribute. Let's do it for both classes

app/main.py
```diff
@@ -1,8 +1,83 @@
...
-    pass
+    def __init__(
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = (self.validate_distance
+                                          (distance_from_city_center))
```
MNDonut: A good practice is to move method and parameter on the next line together. fix it everywhere

MNDonut (changes requested):

Oleksl888 (approved):
Superstar!

PR #1398, opened 2024-07-22
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1,8 +1,52 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: In the requirements these attributes have a range of values they can accept. Let's check values and if they are incorrect, set some default value and is correct

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: In the requirements these attributes have a range of values they can accept. Let's check values and if they are incorrect, set some default value and is correct

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+            if car.clean_mark < self.clean_power:
+                total_cost += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(total_cost, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class
+                 * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
```
MNDonut: let's always follow the first approach from the checklist

MNDonut (changes requested):

Oleksl888 (approved):
Good job!

PR #1399, opened 2024-07-22
Washing Station

yashkunn:


app/main.py
```diff
@@ -1,8 +1,67 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List) -> float:
```
Oleksl888: you can be more specific about list contents

app/main.py
```diff
@@ -1,8 +1,67 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income = self.wash_single_car(car)
+                total_income += income
+        return round(total_income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if not isinstance(car, Car):
+            raise ValueError("The argument must be a Car instance")
+        if car.clean_mark >= self.clean_power:
```
Oleksl888: Do we have a requirement to return price 0 when clean_mark doesnt match clean_power?

app/main.py
```diff
@@ -1,8 +1,67 @@
...
+            return 0.0
+        washing_price = (
```
Oleksl888: You don't need to create a variable here since you not going to reuse it anywhere

app/main.py
```diff
@@ -1,8 +1,67 @@
...
+        return 0.0
+
+    def rate_service(self, new_rating: float) -> None:
+        if not (1.0 <= new_rating <= 5.0):
+            raise ValueError("new_rating must be between 1.0 and 5.0")
+
+        total_ratings = self.average_rating * self.count_of_ratings
+        new_total_ratings = total_ratings + new_rating
+        self.count_of_ratings += 1
+        self.average_rating = round(
```
Oleksl888, line 66: you can calculate new average rating in one statement without creating extra variables

app/main.py
```diff
@@ -1,8 +1,67 @@
...
+
+    def wash_single_car(self, car: Car) -> float:
+        if not isinstance(car, Car):
+            raise ValueError("The argument must be a Car instance")
+
+        if car.clean_mark < self.clean_power:
+            price = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return price
+        return 0.0
```
Oleksl888: there is no requirement to return 0 price when clean mark does not meet clean power

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,62 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
MNDonut: The best practice is to start parameters from the next line after method name

app/main.py
```diff
@@ -1,8 +1,62 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
```
MNDonut, line 24: here you did correctly :)

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: these values have specific range of correct values, let's check if it's correct. If not -> set some default correct value

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: these values have specific range of correct values, let's check if it's correct. If not -> set some default correct value

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        total_income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total_income += self.wash_single_car(car)
```
MNDonut: you can make list comprehension here

MNDonut (changes requested):

app/main.py
Oleksl888: Please set the default value instead of raising exception

app/main.py
Oleksl888: distance to city center can be from 1.0 to 10.0 as per task definition

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,69 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+
+        self.comfort_class = max(1, min(comfort_class, 7))
```
Oleksl888, line 12: This is clever :+1:

Oleksl888 (approved):
Well done!

PR #1400, opened 2024-07-22
implement for py-car-wash-station

Maksym-Turenko:


app/main.py
```diff
@@ -1,8 +1,47 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
MNDonut: Let's always follow the first approach from the checklist

app/main.py
```diff
@@ -1,8 +1,47 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int
+                 ) -> None:
```
MNDonut: Let's always follow the first approach from the checklist

app/main.py
```diff
@@ -1,8 +1,47 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: these attributes can have only specific values, let's check them and if they are incorrect, set default correct value

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int
+                 ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: these attributes can have only specific values, let's check them and if they are incorrect, set default correct value

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = ((car.comfort_class
+                  * (self.clean_power - car.clean_mark)
+                  * self.average_rating)
+                 / self.distance_from_city_center)
```
MNDonut: these attributes can have only specific values, let's check them and if they are incorrect, set default correct value

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,83 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center \
```
Oleksl888: As per styling guidelines we don't use `\` for line breakers - use `()` on the expression instead

app/main.py
```diff
@@ -1,8 +1,83 @@
...
+            return average_rating
+        return 3
+
+    @staticmethod
+    def validate_count_of_ratings(count_of_ratings: int) -> int:
+        if count_of_ratings >= 0:
+            return count_of_ratings
+        return 0
+
+    def serve_cars(self, cars: list) -> float:
```
Oleksl888: Can you speicify what data type is expected inside of the cars list?

app/main.py
```diff
@@ -1,8 +1,83 @@
...
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = ((car.comfort_class
```
Oleksl888: No need to store result of calculation in variable since you are not going to reuse it. You can return result straight:
```
return round(
            car.comfort_class
            ...

File: app/main.py [hunk 9]
```diff
@@ -1,8 +1,83 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        total_rating += new_rating
+        self.average_rating = round(total_rating / self.count_of_ratings, 1)
```
Oleksl888: Can you rewrite this formula so you don't have to use total_rating variable ?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,88 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
MNDonut: The best approach when parameters don't fit in 1 line is to move them on the next lines after method name

app/main.py
```diff
@@ -1,8 +1,88 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int) -> None:
```
MNDonut: The best approach when parameters don't fit in 1 line is to move them on the next lines after method name

app/main.py
```diff
@@ -1,8 +1,88 @@
...
+    def validate_comfort_class(comfort_class: int) -> int:
+        if 1 <= comfort_class <= 7:
+            return comfort_class
+        return 3
+
+    @staticmethod
+    def validate_clean_mark(clean_mark: int) -> int:
+        if 1 <= clean_mark <= 10:
+            return clean_mark
+        return 5
```
MNDonut: Basically you use methods wrong. You can note that all of them have the same pattern. It means you can create 1 method that accepts all parameters and return correct values. In this way you can create 1 method instead of 20 when you have 20 parameters that should be checked

app/main.py
```diff
@@ -1,8 +1,88 @@
...
+    def validate_average_rating(average_rating: int) -> int:
+        if 1 <= average_rating <= 5:
+            return average_rating
+        return 3
+
+    @staticmethod
+    def validate_count_of_ratings(count_of_ratings: int) -> int:
+        if count_of_ratings >= 0:
+            return count_of_ratings
+        return 0
```
MNDonut: Basically you use methods wrong. You can note that all of them have the same pattern. It means you can create 1 method that accepts all parameters and return correct values. In this way you can create 1 method instead of 20 when you have 20 parameters that should be checked

app/main.py
MNDonut: Also no need to use staticmethod because you call this method using self

app/main.py
```diff
@@ -1,8 +1,88 @@
...
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(((car.comfort_class
+                       * (self.clean_power - car.clean_mark)
+                       * self.average_rating)
+                      / self.distance_from_city_center), 1)
```
```
variable = (
    a + b
    * 2
    * a
)
```

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str) -> None:
```
Oleksl888, line 9: This is still incorrect styling. Should be either/or:
```
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str,
    ) -> None:
    def __init__(self,
                comfort_class: int,
                clean_mark: int,
                brand: str) -> None:

File: app/main.py [hunk 16]
```diff
@@ -1,8 +1,80 @@
...
+        self.comfort_class = self.validate_parameter(comfort_class, 1, 7, 3)
+        self.clean_mark = self.validate_parameter(clean_mark, 1, 10, 5)
+        self.brand = brand
+
+    @staticmethod
+    def validate_parameter(
+            value: int,
+            min_value: int,
+            max_value: int,
+            default: int) -> int:
```
Oleksl888, line 19: same here - please fix as per styling guidelines

app/main.py
```diff
@@ -1,8 +1,80 @@
...
+    def serve_cars(self, cars: List[Car]) -> float:
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = ((
```
Oleksl888, line 64: This haven't been refactored - you don't need to store this variable

Oleksl888 (approved):
Approved - but please check latest comments and fix accordingly

PR #1401, opened 2024-07-22
implement py-car-wash-station

ValeriiaKyr:


app/main.py
```diff
@@ -1,8 +1,38 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```
Oleksl888: Writing each parameter from a new line is a best practice

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
```
Oleksl888: Same as above - too many params in one line

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Oleksl888: You can be more specific about data type inside of the list

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        price = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return round(price, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark) *
```
Oleksl888: Break lines here as per below - right now your code doesnt pass flake8 checks:
```
round(

    ...,

    1

)

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/main.py (hunk 1 above)
Comment (MNDonut) [reply app/main.py]: @ValeriiaKyr this comment is not fixed. Parameters should start from the next line after method name

File: app/main.py (hunk 2 above)
Comment (MNDonut) [reply app/main.py]: @ValeriiaKyr still not fixed

File: app/main.py [hunk 5]
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: In the requirements these attributes can have only specific range of values. Let's check them and if they are incorrect - set some default correct value

app/main.py
```diff
@@ -1,8 +1,46 @@
...
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: In the requirements these attributes can have only specific range of values. Let's check them and if they are incorrect - set some default correct value

MNDonut (changes requested):

MNDonut:
Also fixed the comments above where I mentioned you

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = (
+            distance_from_city_center
+            if 1.0 <= distance_from_city_center <= 10.0
+            else 1.0
+        )
+        self.clean_power = clean_power
+        self.average_rating = average_rating if 1.0 <= average_rating <= 5.0 \
```
Oleksl888: As per styling guidelines we  don't use backslash for breaking lines. You can do the same way you did it on line 17-21

app/main.py
Oleksl888: Please add data types that will be expected inside of the list

app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class if 1 <= comfort_class <= 7 else 1
```
Oleksl888, line 9: Good job on adding data validation. Later in the course we will learn how to make it even better :+1:

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job!

PR #1402, opened 2024-07-22
Solution of py-car-wash-station

AndriiHamasa:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: int, clean_power: int,
```
Oleksl888: Using new line for each parameter is a best practice

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Oleksl888: you can be more specific about data type inside the list

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        served_car = [car for car in cars if car.clean_mark < self.clean_power]
```
Oleksl888: Why do you need this if you have a check for clean_mark vs clean_power below?

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        income = 0
+        for car in served_car:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        difference = self.clean_power - car.clean_mark
```
Oleksl888: No need to store extra variables here since you don't reuse them - you can return the result straight away

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        fraction = self.average_rating / self.distance_from_city_center
+        return round(car.comfort_class * difference * fraction, 1)
+
+    def wash_single_car(self, car: Car) -> None:
```
Oleksl888, line 40: As per task requirements this method should check clean_mark vs station clean_power:

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.count_of_ratings += 1
+        numerator = self.average_rating * (self.count_of_ratings - 1) + rate
```
Oleksl888: No need to create extra variable that you will not reuse, if the formula gets too long break lines so your code follows styling convention

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done sir!

PR #1403, opened 2024-07-23
'Solution'

KrenkySpenky:


app/main.py
```diff
@@ -1,8 +1,50 @@
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply the same for the `CarWashStation` class as well

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+                car.comfort_class
+                * (self.clean_power
+                   - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            print(f"Washing {car.brand} at the station.")
```
Dimosphen1: There is no need to add `print` statements since they are not mentioned in the task description. Remove them

Dimosphen1 (changes requested):
Make sure that all tests pass. Several changes were requested

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+                 clean_power: float,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
+        revenue = sum(self.calculate_washing_price(car) for car in car_list)
```
pavlopro: Combi

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+
+        for car in car_list:
```
pavlopro, line 25: Include calculating `calculate_washing_price` into this cycle

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+
+        return round(revenue, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if self.is_can_wash(car):
+            return (car.comfort_class
+                    * (self.clean_power - car.clean_mark)
+                    * self.average_rating
+                    / self.distance_from_city_center)
+        else:
```
pavlopro: else operator is redundant here. just return 0

pavlopro (changes requested):

app/main.py
KrenkySpenky: не понял этот коментарий

pavlopro (approved):
Good Job!

PR #1404, opened 2024-07-23
Solved

Alexandr-Politov:


app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int = 1,
+                 clean_mark: int = 1,
+                 brand: str = "") -> None:
```
MNDonut: the best practice is to start parameters from the next line

app/main.py
```diff
@@ -1,8 +1,46 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float = 1.0,
+                 clean_power: int = 1,
+                 average_rating: float = 1.0,
+                 count_of_ratings: int = 0):
```
MNDonut: the best practice is to start parameters from the next line

app/main.py
```diff
@@ -1,8 +1,46 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int = 1,
+                 clean_mark: int = 1,
+                 brand: str = "") -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: these values have specific range of correct values. Let's check if they are correct, if not -> set some default correct value

app/main.py
```diff
@@ -1,8 +1,46 @@
...
-    pass
+    def __init__(self,
+                 distance_from_city_center: float = 1.0,
+                 clean_power: int = 1,
+                 average_rating: float = 1.0,
+                 count_of_ratings: int = 0):
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: these values have specific range of correct values. Let's check if they are correct, if not -> set some default correct value

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        washing_price = ((car.comfort_class 
+                          * (self.clean_power - car.clean_mark) 
+                          * self.average_rating) 
+                         / self.distance_from_city_center)
```
MNDonut: the best practice is to start calculations from the next line when you use ()

MNDonut (changes requested):

MNDonut:
And tests failed :)

.github/workflows/test.yml
Oleksl888, line 1: Please roll back changes from this file

app/main.py
```diff
@@ -1,8 +1,56 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int = 1,
```
Oleksl888: There was no requirement to set default values for attributes.

app/main.py
```diff
@@ -1,8 +1,56 @@
...
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float = 1.0,
+            clean_power: int = 1,
+            average_rating: float = 1.0,
+            count_of_ratings: int = 0
+    ) -> None:
+
+        self.distance_from_city_center = distance_from_city_center \
```
Oleksl888: As per our style guidelines we don't use `\` for line breakers, here you can use `()`  instead

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int) -> float:
+        old_rating = self.average_rating * self.count_of_ratings
+        new_rating = old_rating + rating
+        self.count_of_ratings += 1
+        self.average_rating = round(new_rating / self.count_of_ratings, 1)
+        return self.average_rating
```
Oleksl888: can you return average rating without creating extra variables?

Oleksl888 (changes requested):

app/main.py
Oleksl888: You normally create a variable when you are going to reuse it's value, so you don't have to do recalculate it. In code above `old_rating` variable is unnecessary. So you update your value like so
```
self.average_rating = round(((self.average_rating
                                      * self.count_of_ratings
                                      + rating)
                                     / ...

And by the way there is no requirement to return average rating in this method - you just need to update attribute in class

File: app/main.py [hunk 9]
```diff
@@ -1,8 +1,62 @@
...
+            else 1.0
+        )
+        self.clean_power = clean_power
+        self.average_rating = (
+            average_rating if 1.0 <= average_rating <= 5.0 else 1.0
+        )
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        washing_price = (
```
Oleksl888: You can return result without using an extra variable:
```
return round(
            car.comfort_class
            ....

Oleksl888 (changes requested):

MNDonut (approved):

PR #1405, opened 2024-07-23
'Solution'

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -5,15 +5,10 @@ class Car:
             clean_mark: int,
             brand: str
     ) -> None:
-        # I use this order in initializate default value for not using "else"
-        self.comfort_class = comfort_class
-        if not 1 <= comfort_class <= 7:
-            self.comfort_class = 1
-
-        self.clean_mark = clean_mark
-        if not 1 <= clean_mark <= 10:
-            self.clean_mark = 1
-
+        self.comfort_class = comfort_class \
+            if 1 <= comfort_class <= 7 else 1
+        self.clean_mark = clean_mark \
+            if 1 <= clean_mark <= 10 else 1
         self.brand = brand
 
 
@@ -25,32 +20,26 @@ class CarWashStation:
             average_rating: float,
             count_of_ratings: int
     ) -> None:
-        self.distance_from_city_center = distance_from_city_center
-        if not 1.0 <= distance_from_city_center <= 10.0:
-            self.distance_from_city_center = 1.0
-
-        self.clean_power = clean_power
-        if not 1 <= clean_power <= 10:
-            self.clean_power = 1
-
-        self.average_rating = average_rating
-        if not 1.0 <= average_rating <= 5.0:
-            self.average_rating = 1.0
-
+        self.distance_from_city_center = distance_from_city_center \
+            if 1.0 <= distance_from_city_center <= 10.0 else 1.0
+        self.clean_power = clean_power \
+            if 1 <= clean_power <= 10 else 1
+        self.average_rating = average_rating \
+            if 1.0 <= average_rating <= 5.0 else 1.0
         self.count_of_ratings = count_of_ratings
 
     def wash_single_car(self, car: Car) -> None:
-        car.clean_mark = self.clean_power
+        if self.clean_power > car.clean_mark:
+            clean_mark = self.clean_power
+            car.clean_mark = clean_mark
 
     def serve_cars(self, cars: list) -> float:
-        # sorted cars, who car_station can wash
-        washes = [car for car in cars if car.clean_mark < self.clean_power]
-        # calculate income
-        income = [self.calculate_washing_price(car) for car in washes]
-        # washing sorted car
-        for car in washes:
-            self.wash_single_car(car)
-        return round(sum(income), 1)
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
 
     def calculate_washing_price(self, car: Car) -> float:
         return round(
@@ -59,8 +48,8 @@ class CarWashStation:
         )
 
     def rate_service(self, rate: int) -> None:
-        self.average_rating = (
-            round((self.average_rating * self.count_of_ratings + rate)
-                  / (self.count_of_ratings + 1), 1)
-        )
         self.count_of_ratings += 1
+        self.average_rating = round(
+            (self.average_rating * (self.count_of_ratings - 1) + rate)
+            / self.count_of_ratings, 1
+        )
--- a/app/main.py
+++ b/app/main.py
@@ -5,8 +5,15 @@ class Car:
             clean_mark: int,
             brand: str
     ) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,48 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: these parameters can have specific values, let's check them and if not correct - set some default correct value

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
```
MNDonut: these parameters can have specific values, let's check them and if not correct - set some default correct value

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        # calculate income
+        income = [self.calculate_washing_price(car) for car in washes]
+        # washing sorted car
+        for car in washes:
+            self.wash_single_car(car)
+        return round(sum(income), 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
```
MNDonut: it's the best practice to move everything inside the first brackets on the next lines

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,66 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        # I use this order in initializate default value for not using "else"
```
Oleksl888: Try to use ternary operator here instead. Also remove the above comment in final commit

app/main.py
```diff
@@ -1,8 +1,66 @@
...
+        if not 1 <= clean_power <= 10:
+            self.clean_power = 1
+
+        self.average_rating = average_rating
+        if not 1.0 <= average_rating <= 5.0:
+            self.average_rating = 1.0
+
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car) -> None:
```
Oleksl888, line 38: Task definition requires to check this condition before:

app/main.py
```diff
@@ -1,8 +1,66 @@
...
+        car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
+        # sorted cars, who car_station can wash
+        washes = [car for car in cars if car.clean_mark < self.clean_power]
```
Oleksl888: By making this check you add extra complexity to your code. There is no need to make a separate iteration to filter cars - you can do one iteration, check the condition and execute rest of code if it is met.

app/main.py
```diff
@@ -1,8 +1,66 @@
...
+        # calculate income
+        income = [self.calculate_washing_price(car) for car in washes]
```
Oleksl888: Same here - no need to make an extra iteration to calculate income - all can be done in one iteration:
```
for car in cars:
   check condition
   if yes -> calculate income and wash the car
   if no -> do nothing

File: app/main.py [hunk 8]
```diff
@@ -1,8 +1,66 @@
...
+        return round(sum(income), 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center, 1
+        )
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = (
```
Oleksl888: here you can do same as you did on line 56:
```
self.average_rating = round(
...
)

File: app/main.py [hunk 9]
```diff
@@ -1,8 +1,66 @@
...
+
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: list) -> float:
+        # sorted cars, who car_station can wash
+        washes = [car for car in cars if car.clean_mark < self.clean_power]
+        # calculate income
```
Oleksl888: remove self explanatory comments

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,55 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = comfort_class \
```
Oleksl888: As per styling guide we do not use `\` for line breakers, here you can put your expression in `()` instead

app/main.py
```diff
@@ -1,8 +1,55 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center \
```
Oleksl888: Refactor so you don't use `\` for breaking lines

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,57 @@
...
+    ) -> None:
+        self.distance_from_city_center = (
+            distance_from_city_center
+            if 1.0 <= distance_from_city_center <= 10.0 else 1.0
+        )
+        self.clean_power = (clean_power
+                            if 1 <= clean_power <= 10 else 1)
+        self.average_rating = (average_rating
+                               if 1.0 <= average_rating <= 5.0 else 1.0)
+        self.count_of_ratings = count_of_ratings
```
MNDonut: we never leave a part of instruction on the same lvl that the opening parenthesis. You should move the whole structure on te next line

app/main.py
```diff
@@ -1,8 +1,57 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            clean_mark = self.clean_power
+            car.clean_mark = clean_mark
```
MNDonut: You don't need this variable here, just to direct assignment

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,59 @@
...
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
+        self.comfort_class = (comfort_class
+                              if 1 <= comfort_class <= 7 else 1)
+        self.clean_mark = (clean_mark
+                           if 1 <= clean_mark <= 10 else 1)
```
MNDonut: still not fixed

MNDonut (changes requested):

MNDonut (approved):

PR #1406, opened 2024-07-23
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: In the requirements these attributes can accept only specific values. Let's check them and if value is incorrect, set some default correct value

app/main.py
```diff
@@ -1,8 +1,41 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: In the requirements these attributes can accept only specific values. Let's check them and if value is incorrect, set some default correct value

app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
MNDonut: We don't format parameters in this way. The best approach is: parameters start from the next line and 1 line - 1 parameter

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
MNDonut: We don't format parameters in this way. The best approach is: parameters start from the next line and 1 line - 1 parameter

MNDonut (changes requested):

MNDonut (approved):

PR #1407, opened 2024-07-23
Solution

dpiuro:


app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+        self, comfort_class: int, clean_mark: int, brand: str
```
Oleksl888: When you have multiple params it is a best practice to have each parameter on new line

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+        self, distance_from_city_center: float, clean_power: int,
```
Oleksl888: Same here  - each param must start on a new line

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (
```
Oleksl888: No need to store result of calculation in variable since you are not going to reuse it. You can straight return:
```
return round(
            car.comfort_class
            ...

File: app/main.py [hunk 4]
```diff
@@ -1,8 +1,44 @@
...
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, upd_rating: int) -> None:
+        rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round(
```
Oleksl888, line 63: Can you rewrite this formula so you don't have to use rating variable ?

Oleksl888 (changes requested):

app/main.py
dpiuro: Hi! Please, check only "Fixed 4 comments"

app/main.py
```diff
@@ -1,8 +1,52 @@
...
-    pass
+    def __init__(
+        self,
+        comfort_class: int,
+        clean_mark: int,
+        brand: str
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
```
MNDonut, line 15: Note that some parameters can accept only specific values

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut, line 38: Note that some parameters can accept only specific values

MNDonut (changes requested):

MNDonut (approved):

PR #1408, opened 2024-07-23
'Solution'

ABaidin:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int = 7,
+            clean_mark: int = 1,
+            brand: str = "some_car"
+    ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
```
MNDonut: In the requirements these attributes can accept only specific values. Let's check them and if value is incorrect, set some default correct value

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+            self,
+            distance_from_city_center: float = 1.0,
+            clean_power: int = 10,
+            average_rating: float = 5.0,
+            count_of_ratings: int = 100
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: In the requirements these attributes can accept only specific values. Let's check them and if value is incorrect, set some default correct value

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rating: int | float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round((total_rating + rating)
+                                    / self.count_of_ratings, 1)
```
```
variable = (
    a + b
    * 2
    * a
)
```

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
```
```
variable = (
    a + b
    * 2
    * a
)
```

MNDonut (changes requested):

ABaidin:
I'm not sure if I implemented calculation adjustments right, because flake8 was cursing on me all the time, but I managed to write code the way it passes pytest and flake8. Hope it looks good enough and I did what you wanted from me :)
Also, the conditions and default values that I wrote seems a little bit sketchy to me. Was I supposed to use "If: else:", or this way is also viable?

app/main.py
```diff
@@ -5,22 +5,32 @@ def __init__(
             clean_mark: int = 1,
             brand: str = "some_car"
     ) -> None:
-        self.comfort_class = comfort_class
-        self.clean_mark = clean_mark
+        if comfort_class in range(1, 7 + 1):
+            self.comfort_class = comfort_class
+
+        if clean_mark in range(1, 10 + 1):
+            self.clean_mark = clean_mark
```
MNDonut: what if I pass -100?
it means you'll not set self.comfort_class at all and the same for other attributes
basically you can still use if statement or even create a separate function, but if statement should be correct and value should be set as well :)

app/main.py
```diff
@@ -42,5 +56,8 @@ def wash_single_car(self, car: Car) -> None:
     def rate_service(self, rating: int | float) -> None:
         total_rating = self.average_rating * self.count_of_ratings
         self.count_of_ratings += 1
-        self.average_rating = round((total_rating + rating)
-                                    / self.count_of_ratings, 1)
+        self.average_rating = (round(
```
MNDonut: round should be on the next line
the general rule is: content inside brackets should be on other lines than these brackets

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,76 @@
...
+        price = (
+            (self.clean_power - car.clean_mark)
+            * car.comfort_class
+            * self.average_rating
+            / self.distance_from_city_center
+        )
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Oleksl888: missing condition from p3 in task definition

app/main.py
```diff
@@ -1,8 +1,76 @@
...
+            self.average_rating = average_rating
+        else:
+            self.average_rating = 3.5
+
+        if count_of_ratings > 0:
+            self.count_of_ratings = count_of_ratings
+        else:
+            self.count_of_ratings = 10
+
+    def serve_cars(self, cars: list) -> float:
```
Oleksl888: Specify cars list contents data type

app/main.py
```diff
@@ -1,8 +1,76 @@
...
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (
```
Oleksl888: No need to store result in variable since you are not going to reuse this value

Oleksl888 (changes requested):

app/main.py
ABaidin: "wash_single_car" called only inside "serve_cars" where this condition needs to be checked anyway.
Do I still need to add this condition inside "wash_single_car" method?

Oleksl888: yes. It is written in task requirements + this method potentially can be called outside other method

LiudmylaKulinchenko (approved):
Well done 👏

PR #1409, opened 2024-07-23
Solution

pashawarganov:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+    return num
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
MNDonut: The best approach in this case is to move all parameters on the next lines after method name

app/main.py
```diff
@@ -1,8 +1,51 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
MNDonut: The best approach in this case is to move all parameters on the next lines after method name

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round((self.average_rating
+                                     * self.count_of_ratings + rate)
+                                    / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
```
variable = (
    a + b
    * 2
    * a
)
```

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,62 @@
+from __future__ import annotations
+
+
+def limitation(
+        num: int | float,
+        min_lim: int | float,
+        max_lim: int | float) -> int | float:
```
MNDonut: the last bracket + annotation should be on a separate line

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str) -> None:
```
MNDonut: same here

app/main.py
```diff
@@ -1,8 +1,62 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int) -> None:
```
MNDonut: same here

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+        for index, car in enumerate(cars):
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(cars[index])
+                self.wash_single_car(cars[index])
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_center, 1)
```
MNDonut: 1 and ) should be on separate lines as 1 is parameter and the last bracket is always on a separate line

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        self.average_rating = round(
+            (self.average_rating
+             * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1), 1
+        )
```
MNDonut: move 1 on the next line

MNDonut (changes requested):

MNDonut (approved):

PR #1411, opened 2024-07-23
Task 5

Poznanskyi:


app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str)\
```
Oleksl888: When you have multiple params it is a best practice to have each parameter on new line. Also we don't use `\` for line breakers as per styling guidelines

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+        self, distance_from_city_center: float, clean_power: int,
```
Oleksl888: Same as above - each param must be on separate line

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        average_rating: float, count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Oleksl888: Can you specify data type for list contents?

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        cost = (
```
Oleksl888: No need to store result of calculation in variable since you are not going to reuse it.

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round(
```
Oleksl888, line 71: Same as above - you can rewrite this formula so you don't have to create total_rating variable

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,50 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
MNDonut: The best approach is to move these parameters on the next lines after function name

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
```
MNDonut, line 19: Note that some parameters can accept only specific values

app/main.py
```diff
@@ -1,8 +1,50 @@
...
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut, line 45: Note that some parameters can accept only specific values

app/main.py
```diff
@@ -1,8 +1,50 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
MNDonut: The best approach is to move these parameters on the next lines after function name

MNDonut (changes requested):

app/main.py
```diff
@@ -2,21 +2,43 @@
...
+            count_of_ratings: int
+    ) -> None:
+        if distance_from_city_center <= 0:
+            raise ValueError("Must be greater than 0")
+
+        if clean_power not in range(0, 11):
+            raise ValueError("Must be between 0 and 10")
+
+        if not (0.0 <= average_rating <= 5.0):
+            raise ValueError("Must be between 0.0 and 5.0")
```
MNDonut, line 37: let's follow the same approach everywhere. Either double comparison or range

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,75 @@
...
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        if distance_from_city_center <= 0:
+            raise ValueError("Must be greater than 0")
+
+        if clean_power not in range(0, 11):
+            raise ValueError("Must be between 0 and 10")
+
+        if not (0.0 <= average_rating <= 5.0):
```
MNDonut, line 36: still not fixed. you use here double comparison, but for other ifs use range

app/main.py
```diff
@@ -1,8 +1,75 @@
...
+            * self.average_rating / self.distance_from_city_center,
+            1
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        if not (0.0 <= new_rating <= 5.0):
```
MNDonut, line 67: the same here

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Good job 👏

PR #1414, opened 2024-07-24
Solution

vStryhun:


app/main.py
```diff
@@ -1,8 +1,63 @@
...
+    def __init__(
+        self,
+        comfort_class: int,
+        clean_mark: int,
+        brand: str
+    ) -> None:
+        if not (1 <= comfort_class <= 7):
+            print("Error")
+        if not (1 <= clean_mark <= 10):
+            print("Error")
```
MNDonut: What if I add this code at the end?
```
wrong_car = Car(-100, -100, "name")
Car.serve_cars([wrong_car])
```
It'll print this text, but the program will continue to work with incorrect data, thus you should either set correct data or stop program

app/main.py
```diff
@@ -1,8 +1,63 @@
...
+        clean_power: bool,
+        average_rating: float,
+        count_of_ratings: int,
+    ) -> None:
+        if not (1.0 <= distance_from_city_center <= 10.0):
+            print("Error")
+        if not (1 <= clean_power <= 10):
+            print("Error")
+        if not (1.0 <= average_rating <= 5.0):
+            print("Error")
```
MNDonut: What if I add this code at the end?
```
wrong_car = Car(-100, -100, "name")
Car.serve_cars([wrong_car])
```
It'll print this text, but the program will continue to work with incorrect data, thus you should either set correct data or stop program

MNDonut (changes requested):

app/main.py
```diff
@@ -6,9 +6,9 @@ def __init__(
         brand: str
     ) -> None:
         if not (1 <= comfort_class <= 7):
-            print("Error")
+            raise ValueError("comfort_class must be between 1 and 7")
         if not (1 <= clean_mark <= 10):
-            print("Error")
+            raise ValueError("clean_mark must be between 1 and 10")
```
MNDonut, line 11: yeah, or just set some default value, but that's okay :)

MNDonut (changes requested):
Tests failed :)

PR #1441, opened 2024-08-10
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1,8 +1,37 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
LiudmylaKulinchenko: It's better to make one line - one parameter

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 💪

app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
LiudmylaKulinchenko, line 5: Also, just for you to note, it's more common to make an indentation after parentheses, like in this [checklist ](https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md) first Good example. So your option is ok, but better to use other one

LiudmylaKulinchenko (approved):
Well done 💪

PR #1458, opened 2024-08-20
Solution

KostiantinPastukh:


app/main.py
```diff
@@ -1,8 +1,53 @@
...
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        total_income = 0.0
+        cars_list = [car for car in cars if car.clean_mark < self.clean_power]
```
pavlopro: No need allocate memory for this list, just use regular iteration and check that car.clean_mark < self.clean_power

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        for car in cars_list:
+            total_income += self.calculate_washing_price(car)
+            self.wash_single_car(car)
+        return round(total_income, 1)
```
pavlopro: No need to have this `round `here

app/main.py
```diff
@@ -1,8 +1,53 @@
...
+
+    def rate_service(self, new_rating: float) -> None:
+        total = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round(
+            (
+                total + new_rating
+            ) / self.count_of_ratings,
+            1
+        )
```
pavlopro, line 52: Please make this more prettier to improve readability

pavlopro (changes requested):

app/main.py
KostiantinPastukh: @pavlopro check new solution plz

vasylhnatiuk (approved):

PR #1463, opened 2024-08-25
 py-car-wash-station

Mxbely:
Created class Car and CarWashStation
Calculated income from washing cars

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,5 +1,3 @@
-from typing import Any
-
 
 class Car:
     def __init__(self, comfort_class: int,
@@ -20,18 +18,16 @@ class CarWashStation:
         self.average_rating = average_rating
         self.count_of_ratings = count_of_ratings
 
-    def serve_cars(self, cars_list: list) -> Any:
+    def serve_cars(self, cars_list: list) -> int:
         result = 0
         for car in cars_list:
-            if car.clean_mark is None:
-                result += 0
-            elif car.clean_mark < self.clean_power:
-                result += self.wash_single_car(car.clean_mark,
-                                               car.comfort_class)
+            if car.clean_mark < self.clean_power:
+                result += self.calculate_washing_price(car)
+                self.wash_single_car(car.clean_mark)
                 car.clean_mark = self.clean_power
         return result
 
-    def calculate_washing_price(self, car: Car) -> Any:
+    def calculate_washing_price(self, car: Car) -> float:
         calculate = 0
         if car.clean_mark < self.clean_power:
             calculate = (car.comfort_class
@@ -40,17 +36,12 @@ class CarWashStation:
                          / self.distance_from_city_center)
         return round(calculate, 1)
 
-    def wash_single_car(self, car_mark: int, car_class: int) -> Any:
-        calculate = 0
-        if car_mark < self.clean_power:
-            calculate = (car_class * (self.clean_power - car_mark)
-                         * self.average_rating
-                         / self.distance_from_city_center)
-        return round(calculate, 1)
+    def wash_single_car(self, clean_mark: int) -> None:
+        if clean_mark < self.clean_power:
+            clean_mark = self.clean_power
 
-    def rate_service(self, mark: int) -> Any:
+    def rate_service(self, mark: int) -> None:
         self.count_of_ratings += 1
         result = (((self.average_rating * (self.count_of_ratings - 1)) + mark)
                   / self.count_of_ratings)
         self.average_rating = round(result, 1)
-        return self.average_rating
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,56 @@
...
+    def __init__(self, distance_from_city_center: int,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> Any:
```
AnyoneClown: It is a bad practice to specify the `Any` type in annotations. Figure out what you're returning, and if you're not returning anything, return `None`

Change in other lines too

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+        result = 0
+        for car in cars_list:
+            if car.clean_mark is None:
+                result += 0
```
AnyoneClown: There is no need to add 0 to your result, this action will not make any change. Simply skip this condition and only satisfy the condition in `car.clean_mark < self.clean_power` case

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+                         / self.distance_from_city_center)
+        return round(calculate, 1)
+
+    def wash_single_car(self, car_mark: int, car_class: int) -> Any:
+        calculate = 0
+        if car_mark < self.clean_power:
+            calculate = (car_class * (self.clean_power - car_mark)
+                         * self.average_rating
+                         / self.distance_from_city_center)
+        return round(calculate, 1)
```
AnyoneClown: The `wash_single_car` method must perform the logic described in the name of this method. You should not duplicate the logic of several methods, use the `calculate_washing_price` method to calculate

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> Any:
+        result = 0
+        for car in cars_list:
+            if car.clean_mark is None:
+                result += 0
+            elif car.clean_mark < self.clean_power:
+                result += self.wash_single_car(car.clean_mark,
```
AnyoneClown: Result should be counted by `calculate_washing_price` method

app/main.py
AnyoneClown: This func should not returen anything

app/main.py
```diff
@@ -1,8 +1,56 @@
...
+                         * self.average_rating
+                         / self.distance_from_city_center)
+        return round(calculate, 1)
+
+    def rate_service(self, mark: int) -> Any:
+        self.count_of_ratings += 1
+        result = (((self.average_rating * (self.count_of_ratings - 1)) + mark)
+                  / self.count_of_ratings)
+        self.average_rating = round(result, 1)
+        return self.average_rating
```
AnyoneClown: This func should not return anything

AnyoneClown (changes requested):

app/main.py
Mxbely: I thought it, but how i can call function `calculate_washing_price` inside `wash_single_car`, if i don't have parameter for this function?
I can take a access and call `calculate_washing_price` using `@classmethod`, but what else i need to do?

Mxbely: for example, maybe... or something similar...
```
    @classmethod
    def wash_single_car(cls, car_mark: int, clean_power: int) -> None:
        if car_mark < clean_power:
            cls.calculate_washing_price(cls)
```
i don't unerstand, how a can pass parameter class Car to calculate_washing_price into wash_single_car

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list) -> int:
+        result = 0
+        for car in cars_list:
+            if car.clean_mark < self.clean_power:
+                result += self.calculate_washing_price(car)
+                self.wash_single_car(car.clean_mark)
+                car.clean_mark = self.clean_power
```
Y-Havryliv, line 27: this line is redundant, you do this inside the `wash_single_car` method.

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        return result
+
+    def calculate_washing_price(self, car: Car) -> float:
+        calculate = 0
```
Y-Havryliv, line 31: not the best variable name

Y-Havryliv (approved):

PR #1482, opened 2024-09-06
'Solution'

mshmygel:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power
+                        - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
AnyoneClown: There is no point in doing this check if you do it in the method you are using. You repeat this check twice

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #1485, opened 2024-09-06
Solution

Viacheslav417865:


app/main.py
```diff
@@ -1,8 +1,38 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
```suggestion
    def serve_cars(self, cars: list[Car]=) -> float:
```

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                cost = self.calculate_washing_price(car)
+                income += cost
+                self.wash_single_car(car)  # викликаємо метод wash_single_car
```
vsmutok: please remove this comment

vsmutok (changes requested):

app/main.py
Viacheslav417865: If i "Suggested change"  # def serve_cars(self, cars: list[Car]=) -> float: SyntaxError: expected default value expression

PR #1486, opened 2024-09-07
Develop

Viacheslav417865:


Viacheslav417865:
It seems like there's still an issue with the import of serve_cars from the app.main module. The error message indicates that the serve_cars function cannot be imported from app.main.  how to solve this problem for me

PR #1497, opened 2024-09-18
Solution-py-car-wash-station

skyfoxwork:


app/main.py
```diff
@@ -1,8 +1,47 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
Nattalli: bad code style, fix it. Here is example: https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md#code-style

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
Nattalli: You can give more descriptive type annotations

type list can have a lot of different types inside. For example, a list of integers, strings, and Car class instances. So we recommend clarifying the list type: list[int], list[Car], ...

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def wash_single_car(self, car: Car) -> float:
+        return self.calculate_washing_price(car)
+
+    def rate_service(self, grade: float) -> None:
+        up_count_of_ratings = self.count_of_ratings + 1
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + grade)
+            / up_count_of_ratings, 1)
+        self.count_of_ratings = up_count_of_ratings
```
Nattalli:
```python
up_count_of_ratings = self.count_of_ratings + 1
self.count_of_ratings = up_count_of_ratings
 ```
=
```python
self.count_of_ratings += 1
```

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        cost = (car.comfort_class
+                * (self.clean_power - car.clean_mark)
+                * self.average_rating
+                / self.distance_from_city_center)
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        return self.calculate_washing_price(car)
+
+    def rate_service(self, grade: float) -> None:
```
Nattalli: grade is int

Nattalli (changes requested):
comments added

app/main.py
```diff
@@ -1,8 +1,47 @@
...
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int
+                 ) -> None:
```
Nattalli: bad code style, fix it here too. Here is an example: https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md#code-style

Nattalli (changes requested):
one comment left

app/main.py
skyfoxwork: But in my code I had done that:

class Car:
    # write your code here
    pass
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) -> None:

AnyoneClown (approved):
LGTM

PR #1501, opened 2024-09-18
'Solution'

spa7id:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+            car.comfort_class * diff * self.average_rating
+            / self.distance_from_city_center
+        )
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> int:
+        if car.clean_mark < self.clean_power:
+            price = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return price
```
AnyoneClown: A function must execute only one logic. If it is supposed to wash the car, then you should not return a price. Use other function for that

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                total_price += price
+        return round(total_price, 1)
+
+    def calculate_washing_price(self, car: Car) -> int:
+        diff = self.clean_power - car.clean_mark
+        price = (
+            car.comfort_class * diff * self.average_rating
+            / self.distance_from_city_center
+        )
+        return round(price, 1)
```
AnyoneClown: Method `round` returns `float` type. Change your type annotation(In other methods too)

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
+        total_price = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.wash_single_car(car)
+                total_price += price
+        return round(total_price, 1)
```
AnyoneClown: You should return result of calculating price with `calculate_washing_price`, you don't need to `round` it again, because you did this already inside that method

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
Nattalli, line 3:
```suggestion
    def __init__(
        self, 
        comfort_class: int, 
        clean_mark: int,
        brand: str
    ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
Nattalli, line 11:
```suggestion
    def __init__(
         self, 
         distance_from_city_center: float, 
         clean_power: int,
         average_rating: float, 
         count_of_ratings: int
     ) -> None:
```

Nattalli: or
```suggestion
    def __init__(self, 
         distance_from_city_center: float, 
          clean_power: int,
           average_rating: float, 
           count_of_ratings: int) -> None:
```

Nattalli (approved):
not bad, but fix a little code style

PR #1509, opened 2024-09-23
Solution

honelyq:


app/main.py
```diff
@@ -1,8 +1,48 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
pavlopro: Place each argument on the new line.

 def __init__(self,
Comment
Place each argument on the new line.

```
def __init__(
      self,
      comfort_class: int | float,
      clean_mark: int | float,
      brand: str
) -> None:
```

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+                 ) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
```
pavlopro: Do the same as above here as well

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1514, opened 2024-09-24
Solution

VladyslavBon:
Flake8 didn't let to pass the length of my calculations, that's why this strange syntax is there.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -29,7 +29,9 @@ class CarWashStation:
     def serve_cars(self, cars: List[Car]) -> float:
         total_cost = 0
         for car in cars:
-            total_cost += self.wash_single_car(car)
+            if car.clean_mark < self.clean_power:
+                total_cost += self.calculate_washing_price(car)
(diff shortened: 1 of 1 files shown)


skyfoxwork (approved):
I think , good !)

app/main.py
```diff
@@ -1,8 +1,58 @@
...
+            ) / self.distance_from_city_center,
+            1
+        )
+
+    def wash_single_car(self, car: Car) -> float | int:
+        if car.clean_mark < self.clean_power:
+            single_cost = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return single_cost
+        return 0
```
AnyoneClown: What if we want only to wash car, without charging and calculating customer? Your `wash_single_car` method should only clean your car, without calculating the price and returning anything. Call both methods inside `serve_cars`

AnyoneClown (changes requested):

Nattalli (approved):
good job!

PR #1521, opened 2024-09-26
implement car-wash-station

94nj111:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
VladyslavBon: I think should write datail anotation type in (cars: list[?])

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
VladyslavBon, line 37: This condition already used, so you can delite this

VladyslavBon (approved):

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
AnyoneClown: Let's add `if` to check `clean_mark` and `clean_power` before washing car. What if we want to reuse this method in the future? We will add new `if` check again?

AnyoneClown (approved):
Good, fix only 1 small comment

PR #1527, opened 2024-09-27
implemented py-car-wash-station

yuliiapyrih:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+            self.total_price += self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, mark: int) -> None:
+        total_rating = self.average_rating * self.count_of_ratings
+        self.count_of_ratings += 1
+        self.average_rating = round(
+            (total_rating
+                + mark)
+            / self.count_of_ratings, 1)
```
LiudmylaKulinchenko: fix formatting here, you're combining two format options here. First brace starts with a line breaker, but ends with text

LiudmylaKulinchenko (changes requested):

app/main.py
yuliiapyrih: Done

LiudmylaKulinchenko (approved):
Well done 🌷

PR #1529, opened 2024-09-29
'Solution'

AndriiDementiev:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
```
Rabb1t-Artem: move this line to serve_car method

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+            car.clean_mark = self.clean_power
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
+            return round(car.comfort_class
+                         * (self.clean_power
+                            - car.clean_mark)
+                         * self.average_rating
+                         / self.distance_from_city_center, 1)
+        return 0
```
Rabb1t-Artem: del this line/ it's unnecessary.

Rabb1t-Artem (changes requested):

app/main.py
AnyoneClown: Better to keep it here, in case we might need to reuse this method in other places.

For example, if you need to create a new method where you will wash the cars, removing this line would mean you'd have to replicate it inside the new method as well.

AnyoneClown (approved):
Good job, but you still can add that `if check` back in

PR #1533, opened 2024-09-29
Solution

Den-k0:


app/main.py
```diff
@@ -1,8 +1,53 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
```
vladyslav-tmf: You should round the total income to 1 decimal place in the serve_cars method, as the task specifies. Everything else looks good!

vladyslav-tmf (changes requested):

app/main.py
Den-k0: No, that's correct. I already do this in "calculate_washing_price" method

vladyslav-tmf: Actually, it's correct. I didn't realize it at first, sorry. So, great job!

vladyslav-tmf (approved):

app/main.py
Y-Havryliv: test it locally more thoroughly, it's better to round it here as well.
What if calculate_washing_price return `1.1` 3 times?

Y-Havryliv (changes requested):

app/main.py
Den-k0: Really. If calculate_washing_price return `1.1` 3 times, it will be rounding error and result will be `3.3000000000000003`. 🤔

Y-Havryliv (approved):
looks good

vasylhnatiuk (approved):

PR #1534, opened 2024-09-29
Solution

molodsh1y1:


YaroslavBordovoy (approved):
I wouldn't make an additional calculate_base_price method to calculate the price, this can be done in the calculate_washing_price method. Otherwise, everything is fine.

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+        clean_difference = self.clean_power - car.clean_mark
+        base_price = self.calculate_base_price(car)
+
+        return round(clean_difference * base_price, 1)
+
+    def calculate_base_price(self, car: Car) -> float:
+        return (
+            car.comfort_class * self.average_rating
+            / self.distance_from_city_center
+        )
```
AnyoneClown: Implement this logic in one method `calculate_washing_price`

app/main.py
```diff
@@ -1,8 +1,62 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
AnyoneClown: Let's add `if` to check `clean_mark` and `clean_power` before washing car. What if we want to reuse this method in the future? We will add new `if` check again?

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM!

PR #1535, opened 2024-09-29
'Solution'

YaroslavBordovoy:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
molodsh1y1: Add type hints for method arguments for example: List[Car]

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        total_income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                current_price = self.calculate_washing_price(car)
+                self.wash_single_car(car)
+                total_income += current_price
+
+        return round(total_income, 1)
```
molodsh1y1: Rounding the income value may be unnecessary since you are already rounding in calculate_washing_price

molodsh1y1 (changes requested):

molodsh1y1 (approved):
LGTM

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        total_income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                current_price = self.calculate_washing_price(car)
+                self.wash_single_car(car)
+                total_income += current_price
```
AnyoneClown: We don't need to use `current_price` for that. Just add result of `self.calculate_washing_price` to `total_income`

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+        return total_income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
AnyoneClown: Let's add `if` here to check for `clean_power` and `clean_mark` of car.

AnyoneClown (changes requested):

Y-Havryliv (approved):
nice!

PR #1537, opened 2024-09-30
solution

YaYaYaroslav:


sofiiasavkova (approved):
well done :)

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                wash_cost = self.calculate_washing_price(car)
+                income += wash_cost
```
```suggestion
                income += self.calculate_washing_price(car)
```
it's better a bit, and you don't need this new variable

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        )
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: int) -> None:
+        total_ratings = self.average_rating * self.count_of_ratings
+        total_ratings += new_rating
```
AnyoneClown: You can do this in 1 line

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+        self, distance_from_city_center: int, clean_power: int,
+            average_rating: float, count_of_ratings: int
+    ) -> None:
```
AnyoneClown: If you can't include all arguments in 1 line, then start every argument from a new line

```
    def __init__(
        self,
        clean_power: int,
        average_rating: float,
        count_of_ratings: int
    ) -> None:
```

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
AnyoneClown: Add more descriptive type hint for cars. You can update that this list contains `Car` instances

AnyoneClown (changes requested):

app/main.py
AnyoneClown: You didn't fix it. Be more attentive.

app/main.py
AnyoneClown: You didn't fix it.

AnyoneClown (changes requested):

AnyoneClown (approved):
Good, but don't forget to fix flake8

PR #1538, opened 2024-09-30
Solution

sofiiasavkova:


AndriiDementiev (approved):

app/main.py
```diff
@@ -1,8 +1,39 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
```
Y-Havryliv: not the best line break, it's better to do smth like this:
```python
def __init__(
    self, 
    comfort_class: int,
    ...
) -> None:
```

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list[Car]) -> float:
+        total_income = 0
+        for car in cars_list:
+            if car.clean_mark < self.clean_power:
+                wash_price = self.calculate_washing_price(car)
+                total_income += wash_price
```
```suggestion
                total_income += self.calculate_washing_price(car)
```
this would be better, and you don't need this redundant variable

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars_list: list[Car]) -> float:
+        total_income = 0
+        for car in cars_list:
```
```suggestion
        for car in cars:
```
`cars` would be enough, we see that it's a list from type annotations

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
AnyoneClown: Keep the same code style of identation. Reuse indentation logic that you used in `Car's` `__init__` method

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                * self.average_rating / self.distance_from_city_center)
+        return round(cost, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        total_ratings = self.average_rating * self.count_of_ratings
+        total_ratings += new_rating
```
AnyoneClown: We can do this in 1 line

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM!

PR #1545, opened 2024-09-30
Solution

Bondzik-S:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+    def __init__(self,
+                 distance_from_city_senter: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_senter = distance_from_city_senter
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> None:
```
YaroslavBordovoy: The serve_cars method must return a float data type.

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        total_price = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                total_price += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(total_price, 1)
+
```
YaroslavBordovoy, line 33: No need to round the total_price value, you already did that in the calculate_washing_price method.

YaroslavBordovoy (approved):

app/main.py
Y-Havryliv: +, fix it

app/main.py
Y-Havryliv: no, you need to round it here.
What if calculate_washing_price return `1.1` 3 times?

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,8 +1,48 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
+            brand: str) -> None:
```
AnyoneClown: Fix code style. If you start arguments from a new line, then `) -> None` should be on a new line too

```
def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str
) -> None:
```

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_senter: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
AnyoneClown: Use the same code style for identation. Choose one between 2 ways(start from a new line every argument, or this style) and use it everywhere

But you still need to start each argument on the next line, you have two arguments on the same line in line 3

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating / self.distance_from_city_senter, 1
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
AnyoneClown: > wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark;

Add `if` to check for `clean_mark` and `clean_power`. What if someone want to reuse this method? He will need to create `if` to check for this params

AnyoneClown (changes requested):

app/main.py
```diff
@@ -3,17 +3,19 @@ def __init__(
...
 class CarWashStation:
-    def __init__(self,
-                 distance_from_city_senter: float,
-                 clean_power: int, average_rating: float,
-                 count_of_ratings: int) -> None:
+    def __init__(
+            self,
+            distance_from_city_senter: float,
+            clean_power: int, average_rating: float,
+            count_of_ratings: int) -> None:
```
AnyoneClown: `) -> None:` should end on the new line. And you also have 2 arguments on line 17. Start every argument from a new line

```
def __init__(
        self,
        distance_from_city_senter: float,
        clean_power: int, average_rating: float,
        count_of_ratings: int
) -> None:
```

app/main.py
AnyoneClown: Don't forget to fix this

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1549, opened 2024-09-30
Solution

YaroslavK87:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return (
+            (car.comfort_class * (self.clean_power - car.clean_mark)
+             * self.average_rating) / self.distance_from_city_center
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
```
YaYaYaroslav: checking for the difference between “clean_power” and “clean_mark” seems unnecessary, since this is already checked in the “serve_cars” `method

YaYaYaroslav (approved):

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
```
AnyoneClown: Add more descriptive type hint for `cars`. You can update that this list contains `Car` instances

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        total_income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income = self.calculate_washing_price(car)
+                total_income += income
```
AnyoneClown: You don't need to create income variable for that. Just add result of method to `total_income`

app/main.py
```diff
@@ -1,8 +1,50 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self, comfort_class: int,
+            clean_mark: int,
+            brand: str
+    ) -> None:
```
AnyoneClown: Start every argument from a new line. Check other `__init__` method for this mistake

app/main.py
AnyoneClown: Not quite, it's better to leave it, because there may be a case when you need to reuse this method, and there will be no logic to check whether it is worth washing the car or not

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        total_income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
+                total_income += price
```
AnyoneClown: You didn't fix this!

app/main.py
```diff
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
AnyoneClown: Bad identatiton, start every argument from a new line. Fix in other cases too

```
def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str
) -> None:
```

AnyoneClown (changes requested):

Nattalli (approved):
👍🏻

PR #1550, opened 2024-09-30
Solution

Artem773:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
+        income = 0
+        for car in car_list:
+            if car.clean_mark < self.clean_power:
+                self.wash_single_car(car)
```
levseenko: in my opinion, on this line you need to change lines in places with one another because if you do this calculation, the new clean_mark of car will be used to calculate washing_price of car, so calculations might be wrong
```suggestion
                income += self.calculate_washing_price(car)
                self.wash_single_car(car)
```

levseenko (changes requested):
you need to swap lines 26 and 27

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
```
AnyoneClown, line 22: Add more descriptive type hint for `car_list`

AnyoneClown (approved):
Good, but don't forget to fix small type hint :)

PR #1552, opened 2024-10-01
Solution

yulia-pl:


FursykIvan (changes requested):
"All checks have failed
1 failing check"
This message is because you have a line - def __init__(self, comfort_class: int, clean_mark: int, brand: str) -> None: too long. Break it into several lines, a new commit and push again

FursykIvan (approved):

PR #1560, opened 2024-10-07
Solution

KhrypunovMaxim:
Good afternoon! Why do we write all the classes in main? Is it correct to create a separate file for each class?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,66 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
+        if 1 <= comfort_class <= 7:
+            self.comfort_class = comfort_class
+        else:
+            raise ValueError("Comfort class must be between 1 and 7")
+
+        if 1 <= clean_mark <= 10:
+            self.clean_mark = clean_mark
+        else:
+            raise ValueError("Clean mark must be between 1 and 10")
+
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        if 1.0 <= distance_from_city_center <= 10.0:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,66 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
LLkaia: place each argument on new line

app/main.py
```diff
@@ -1,8 +1,66 @@
...
+            raise ValueError("Clean mark must be between 1 and 10")
+
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
```
LLkaia: same here

LLkaia (changes requested):
Several changes were requested. About modules of your project - you are right for case when classes have no relations and also enough big to separate them by putting in different modules. But for our case, it works to have one module.

app/main.py
```diff
@@ -1,8 +1,72 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
LLkaia: Place self on new line as well

app/main.py
```diff
@@ -1,8 +1,72 @@
...
+        else:
+            raise ValueError("Clean mark must be between 1 and 10")
+
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
```
LLkaia: same here

app/main.py
```diff
@@ -1,8 +1,72 @@
...
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
LLkaia: place last bracket and type annotation on new line as well

LLkaia (changes requested):

LLkaia (approved):
Good job!

PR #1612, opened 2024-11-04
Solution

Chukhno-Maksym:


app/main.py
```diff
@@ -1,8 +1,48 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        total_income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                wash_cost = self.calculate_washing_price(car)
+                total_income += wash_cost
```
julia4406, line 26: Looks good, but you can skip variable "wash_cost" and add washing_price to income:

app/main.py
```diff
@@ -1,8 +1,48 @@
...
+
+        wash_price = (car.comfort_class
+                      * (self.clean_power - car.clean_mark)
+                      * (self.average_rating
+                         / self.distance_from_city_center))
+
+        return round(wash_price, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        if car.clean_mark < self.clean_power:
```
julia4406, line 40: You already comparing cars clean_mark with clean_power in func(serve_cars), so here it's unnecessary.

julia4406 (approved):
Great work! ))

PR #1630, opened 2024-11-11
'Solution'

julia4406:


RuslanKravchuk2907 (approved):

app/main.py
```diff
@@ -1,8 +1,79 @@
...
+    comfort_class - comfort class of a car, from 1 to 7
+    clean_mark - car cleanness mark,
+    from very dirty - 1 to absolutely clean - 10
+    brand - brand of the car
+    """
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str
+                 ) -> None:
```
If you want to have another argument on a line with (, then the last “)” should be next to the argument.

def __init__(self,
                 comfort_class: int,
                 clean_mark: int,
                 brand: str) -> None:
                 
                or (my recommendation)
def __init__(
    self,
    comfort_class: int,
    clean_mark: int,
    brand: str
) -> None:

petrykivd: Also, don't forget to fix it in other places

app/main.py
```diff
@@ -1,8 +1,79 @@
...
+        difference = self.clean_power - car.clean_mark
+        multiplication = car.comfort_class * difference * self.average_rating
+        return round(multiplication / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        """
+        should have clean_mark equals wash station's clean_power,
+        if wash_station.clean_power is greater than car.clean_mark
+        """
+        car.clean_mark = self.clean_power
```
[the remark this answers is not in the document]
petrykivd: I think this is a correct comment from the AI, the task states that the check should be in this method. The fact that we check this condition before calling this method is not important, because we can potentially use this method in some other case where checking inside is necessary

petrykivd (changes requested):

L1nk3rrr (approved):
Well done!

PR #1633, opened 2024-11-13
Solution

fredisson11:
Не розумію, чому не проходить тести, щось не так у методі `serve_cars`, але не знаю що саме.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -33,16 +33,18 @@ class CarWashStation:
             if car.clean_mark < self.clean_power
         ]
 
-        for car in cars_for_washing:
-            self.wash_single_car(car)
-
-        return round(
+        income = round(
             sum(
                 [self.calculate_washing_price(car)
                  for car in cars_for_washing]
             ),
             1
         )
+
+        for car in cars_for_washing:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,76 @@
...
+        for car in cars_for_washing:
+            self.wash_single_car(car)
+
+        return round(
+            sum(
+                [self.calculate_washing_price(car)
+                 for car in cars_for_washing]
+            ),
+            1
+        )
```
Alkaponees: You try to use list for summarize value, recommend debug this code locally

Alkaponees (changes requested):
Check my review

app/main.py
fredisson11: Thanks

fredisson11: > You try to use list for summarize value, recommend debug this code locally

The code has long been fixed and is working, it passes all the tests, but the task is displayed as incomplete on the mate platform.

Alkaponees: Try to send review again

Alkaponees (approved):
LGMT!

PR #1670, opened 2024-11-24
Solved task for mentor review with all passed tests and requirements.…

tkudinov:
… I have used (wash) and (for_count) variables to make line shorter according to the flake8 requirements.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,37 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort: int, mark: int, brand: str) -> None:
+        self.comfort_class = comfort
+        self.clean_mark = mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, dist: float, pwr: int, rate: float, count: int) -> None:
+        self.distance_from_city_center = dist
+        self.clean_power = pwr
+        self.average_rating = rate
+        self.count_of_ratings = count
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,37 @@
...
+        self.distance_from_city_center = dist
+        self.clean_power = pwr
+        self.average_rating = rate
+        self.count_of_ratings = count
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                income += round(self.calculate_washing_price(car), 1)
```
Alkaponees: You don't need to use round in this case because your function already return rounded value

app/main.py
```diff
@@ -1,8 +1,37 @@
...
+
+    def wash_single_car(self, car: Car) -> Car:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+            return car
+
+    def rate_service(self, mark: int) -> None:
+        for_count = ((self.average_rating * self.count_of_ratings) + mark)
+        self.average_rating = round(for_count / (self.count_of_ratings + 1), 1)
+        self.count_of_ratings += 1
```
Alkaponees:
```suggestion
       self.count_of_ratings += 1
        self.average_rating = round(
            (((self.average_rating * (self.count_of_ratings - 1)) + mark)
             / self.count_of_ratings), 1)
```

Alkaponees (changes requested):
Check my review

YuriiSmolii (approved):
LGTM

PR #1684, opened 2024-11-29
Solution

ruslanyemets:


app/main.py
```diff
@@ -1,8 +1,51 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self, comfort_class: int, clean_mark: int, brand: str
+    ) -> None:
```
L1nk3rrr: it would be better to make the same style, one attribute per line

L1nk3rrr (changes requested):
Very nice! Good structured code, only 1 minor styling suggestion for you I have

petrykivd (approved):
Nice! 🚀

PR #1687, opened 2024-12-01
py-car-wash-station

SuperKros:


Bleit1 (changes requested):
you have to update your code to complete it(add data types for all data's)

PR #1694, opened 2024-12-01
Implementation of Car and CarWashStation classes

4ikkamony:


app/main.py
```diff
@@ -1,8 +1,52 @@
...
+    def __init__(
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
```
Dogober: You round average_raiting two timse (here and into def rate_service). If i don't mistake there is no need to do one action two times

Dogober (approved):
Beside my little amendment your code is real optimized and cool

petrykivd (approved):

PR #1696, opened 2024-12-01
develop

ihorhalyskiy:
create task Car wash station

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -23,9 +23,10 @@ class CarWashStation:
         self.average_rating = average_rating
         self.count_of_ratings = count_of_ratings
 
-    def serve_cars(self,
-                   cars: list
-                   ) -> float:
+    def serve_cars(
+            self,
+            cars: list
+    ) -> float:
         income = 0
         for car in cars:
             if car.clean_mark < self.clean_power:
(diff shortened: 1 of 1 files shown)


Bulavka-d (approved):

Bulavka-d:
Great job!

app/main.py
```diff
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
```

def init(self,
comfort_class: int,
clean_mark: int,
brand: str) -> None:

            or (my recommendation)
def init(
self,
comfort_class: int,
clean_mark: int,
brand: str
) -> None:

app/main.py
```diff
@@ -1,8 +1,49 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
```
petrykivd: Accordingly, the entire project should have the same code style. Please fix it everywhere

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+        return price
+
+    def rate_service(self, rate: float) -> None:
+        self.average_rating = round(
+            ((self.average_rating * self.count_of_ratings) + rate)
+            / (self.count_of_ratings + 1),
+            1
+        )
+        self.count_of_ratings += 1
```
petrykivd: In order not to perform the same operation (+1) twice, you can move this addition before calculating the average rating.

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        )
+
+    def wash_single_car(self, car: Car) -> float:
+        if car.clean_mark >= self.clean_power:
+            return 0
+
+        price = self.calculate_washing_price(car)
+        car.clean_mark = self.clean_power
+
+        return price
```
petrykivd: According to the task, each method is responsible for its own role. And the method of washing the car should not be responsible for calculating the price, it should only check whether it can wash the car, and if it is possible, then wash the car.

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+            clean_power: int,
+            average_rating: float,
+            count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
```
petrykivd: Please re-read the assignment and method descriptions. They should do exactly what they are intended to do. 
Review the other methods as well

petrykivd (changes requested):

app/main.py
```diff
@@ -1,8 +1,60 @@
...
+                        ) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self,
+                     rating: int
+                     ) -> float:
+        self.average_rating = round(((self.average_rating
+                                      * self.count_of_ratings) + rating)
+                                    / (self.count_of_ratings + 1), 1)
```
L1nk3rrr: Rewrite it somehow for better readability

app/main.py
```diff
@@ -1,8 +1,60 @@
...
+    def calculate_washing_price(self,
+                                car: [Car]
+                                ) -> float:
+        washing_price = round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center, 1
+        )
+        return washing_price
```
L1nk3rrr: Do you need a variable for the washing price? If you want to have it, it looks like `round(washing_price, 1)` would be better than round in your initialization from the reading side

L1nk3rrr (changes requested):
Lets stick to the one styling of declaring methods and class, do the same styling for your functions as you did it at __init__ method.
I mean change
```
    def serve_cars(self,
                   cars: list
                   ) -> float:
```
to 
```
    def serve_cars(
               self,
               cars: list
        ) -> float:
```

petrykivd (approved):
Good job! 👍

PR #1702, opened 2024-12-01
'Solution'

IvankaKuzin:


Sichkaridze (approved):
Great job! 
In serve_cars function you declared total_income as int type (`total_income = 0`), maybe `total_income = 0.0` has sense too, for declaring this as float

num1 = 0.0
num2 = 0
print(type(num1))   #<class 'float'>
print(type(num2))   #<class 'int'>

app/main.py
```diff
@@ -1,8 +1,41 @@
...
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
```

def init(self,
comfort_class: int,
clean_mark: int,
brand: str) -> None:

        or (my recommendation)
def init(
self,
comfort_class: int,
clean_mark: int,
brand: str
) -> None:

petrykivd (approved):
Nice! 👍

PR #1703, opened 2024-12-01
Solution

Sichkaridze:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_possibility = (self.clean_power - car.clean_mark)
+        numerator = car.comfort_class * clean_possibility * self.average_rating
+
+        return round((numerator / self.distance_from_center), 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
```
taraskhom: In the condition of the problem it is said that the variable 'clean_power' should be greater than 'clean_mark', but you compare with '>=' which makes zero profit in function 'calculate_washing_price' (clean_power - clean_mark = 0) and we wash car (spend our soap and water) for zero profit and we even lose money due to it. I approve this task, but change the comparing

taraskhom (approved):
Great job, your code is well structured and very understandable. You have a little mistake, but check my remark, change one thing and move on!

app/main.py
Sichkaridze: Great point, thank you/ I'll change this

taraskhom (approved):
Perfect!

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            transitional_variable = self.clean_power
+            car.clean_mark = transitional_variable
+
+    def rate_service(self, rate: float) -> None:
+        numerator = (self.average_rating * self.count_of_ratings + rate)
+        denominator = (self.count_of_ratings + 1)
+        self.average_rating = round((numerator / denominator), 1)
+        self.count_of_ratings += 1
```
petrykivd, line 51: No need to add 1 twice, just move self.count_of_rating +=1 before calculating the new average rating

petrykivd (approved):
Well done! 😎

PR #1706, opened 2024-12-01
Solution

MindDevastation:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        clean_diff = self.clean_power - car.clean_mark
+        clean_eff = car.comfort_class * clean_diff * self.average_rating
+        inc = clean_eff / self.distance_from_city_center
+        return inc
+
+    def wash_single_car(self, car: Car) -> Car:
+        car.clean_mark = self.clean_power
```
KseniiaLog, line 40: forgot the checking of car.clean_mark < self.clean_power
for future

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+        return car
```
KseniiaLog, line 41: why returning of car, if we need car.clean_mark ?

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def rate_service(self, number: int) -> None:
+        new_total_rating = self.average_rating * self.count_of_ratings + number
+        new_rating = round(
+            new_total_rating / (self.count_of_ratings + 1),
+            1
+        )
+        self.average_rating = new_rating
+        self.count_of_ratings += 1
```
KseniiaLog, line 50: need of return

KseniiaLog (changes requested):
1) forgot the checking of car.clean_mark < self.clean_power
for future
2) why returning of car, if we need car.clean_mark ? in wash_single_car
3) need of return in rate_service

app/main.py
MindDevastation: I'm checking it on row 25.

KseniiaLog (approved):

app/main.py
MindDevastation: We are replasing variable value of car.clean_mark with self.clean_power and have to return class car to use it in the future. If i return only car.clean_mark nothing will work

app/main.py
MindDevastation: The function returns None, so it doesn't need to return anything. I replace the values of the variables self.average_rating and self.count_of_ratings with new values directly incide the function

PR #1708, opened 2024-12-02
Dev

Professor-Douel:


TretiakHelen (changes requested):
calculate_washing_price, wash_single_car, serve_cars and rate_service functions do not belong to the class. You need to add paddings.

Professor-Douel (commented):
was technical problem with tests

PR #1709, opened 2024-12-02
Solution

yelizaveta-zh:


app/main.py
```diff
@@ -1,8 +1,39 @@
...
+
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center:float, clean_power:int, average_rating:float, count_of_ratings:int):
```
maxakim0v: Alright, in this code we had small problem with flake8, ckeck problems with command in your terminal "flake8 app" and solve the problem.

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+
+    def wash_single_car(self, car: 'Car') -> None:
```
maxakim0v: Here's your functiong is working like the task want. Great job.

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: 'Car') -> float:
+        price = (
+            (car.comfort_class * (self.clean_power - car.clean_mark) * self.average_rating) / self.distance_from_city_center
```
maxakim0v: Your faction is not working if destance_from_city_center will be like Zero. Fix this moment.

maxakim0v: You can't divide your formula on zero.

maxakim0v (changes requested):
So, your code is already working! That's a great news, so don't be upset because you have some mistakes. You almost got it, good luck!

app/main.py
yelizaveta-zh: Fixed this problems now in new commit

app/main.py
yelizaveta-zh: But it doesn't cause an error for me when running tests main_test.py. I think it's because self.distance_from_city_center never equals 0 (1.0 to 10.0) from the condition

PR #1722, opened 2024-12-02
Car_Wash_Station_Solution

SummerKatrin:


app/main.py
```diff
@@ -1,8 +1,41 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
```
petrykivd: Fix the code style in all places.
def __init__(
    self,
    comfort_class: int,
    ... ,
    ) -> None:

petrykivd (changes requested):

app/main.py
SummerKatrin: done

petrykivd (approved):
Great! 🚀

PR #1725, opened 2024-12-03
first solution. all tests passed. flake8 fine

Mykola-Osolinskyi:


app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
Nattalli:
```suggestion
    def __init__(self, 
                 comfort_class: int,
                 clean_mark: int,
                 brand: str) -> None:
```

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int,
```
Nattalli: the same here

Nattalli (changes requested):

Nattalli (approved):

PR #1767, opened 2024-12-22
implement Car and CarWashStation classes

yarzhuravlov:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        serving_cars = [car for car in cars if
```
Y-Havryliv: this whole method seems overcomplicated. 
You can just loop through all cars, add money, then wash it.

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                        car.clean_mark < self.clean_power]
+
+        income = sum(
+            [self.calculate_washing_price(car) for car in serving_cars]
+        )
+
+        for car in serving_cars:
+            self.wash_single_car(car)
+
+        return income
```
Y-Havryliv: also it should be rounded

Y-Havryliv (changes requested):

app/main.py
yarzhuravlov: Changes applied, thanks for the feedback

app/main.py
yarzhuravlov: Changes applied, thanks for the feedback

app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
oliasymak24: if all the parameters do not fit on one line, and you move at least something to others, then each one should be on a new line, then readiability will be better

oliasymak24: Also brackets) 
Check everywhere

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
```
oliasymak24: You can use just sum() function here

oliasymak24 (changes requested):

app/main.py
yarzhuravlov: Thanks for the feedback. Changes applied, but I have a question - is it okay if params staying in one line if they can be in this way

```
class Car:
    def __init__(
        self, comfort_class: int, clean_mark: int, brand: str
    ) -> None:
```

Or is it better if in this situation each parameter is on a new line.

app/main.py
yarzhuravlov: Thanks for the feedback. I didn't find a way how to do it with sum() function without iterating over the list twice. Solution with  filtering on the first step was marked as overcomplicated, so now it takes two iterations to use the sum() function.

app/main.py
oliasymak24: If you move brackets then each parameter from new line

oliasymak24 (approved):

PR #1781, opened 2024-12-30
Solve

Baranotik15:


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        if not (1 <= average_rating <= 5):
+            raise ValueError("Average rating must be between 1 and 5.")
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list[Car]) -> float:
+        """Обслуживает список машин и возвращает общую стоимость мойки."""
```
Y-Havryliv: keep your comments in English

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+        total_income = 0
+        if isinstance(list_of_cars, Car):
```
Y-Havryliv, line 36: why do you need this check?
You can simplify your logic a bit. Loop through cars and then add money to income and then wash it.

app/main.py
Y-Havryliv: fix everywhere

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+            return self.wash_single_car(list_of_cars)
+        for car in list_of_cars:
+            total_income += self.wash_single_car(car)
+        return total_income
```
Y-Havryliv: you need to round it here:
`rounded to 1 decimal`

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                  * self.average_rating / self.distance_from_city_center)
+        return round(result, 1)
+
+    def wash_single_car(self, car: Car) -> float:
+        """Моет одну машину, если мощность станции больше
+         текущей чистоты машины."""
+        if self.clean_power > car.clean_mark:
+            result = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return result
```
Y-Havryliv, line 52: should it return anything?

Y-Havryliv (changes requested):

app/main.py
TopalRuslan: I don't really understand about this point

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        self.brand = brand
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None :
```
oliasymak24: Each parameter must be from new line and check your brackets, fix everywhere

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, list_of_cars: list[Car]) -> float:
+        """Maintains a list of cars and returns the
+        total cost of the car wash."""
+        total_income = 0
+        if isinstance(list_of_cars, Car):
+            return self.wash_single_car(list_of_cars)
+        for car in list_of_cars:
+            total_income += self.wash_single_car(car)
```
oliasymak24: you can use sum() function insted of loop

oliasymak24 (changes requested):

app/main.py
TopalRuslan: do we need to do this if we don't have many arguments and they can be on one line?

oliasymak24: Yes, if all the parameters do not fit on one line, and you move at least something to others, then each one should be on a new line, then readiability will be better

oliasymak24 (approved):

PR #1817, opened 2025-01-20
Solution

Slava-Nykonenko:
Classes and methods were created.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int, brand: str) \
```
SerhiyTo: It's best not to use ‘\’ to move data to a new line. The option of transferring the function parameters to a new line is considered the best practice.

SerhiyTo (approved):

PR #1833, opened 2025-01-29
Solution

kbatrakov:


vsmutok (changes requested):
Tests Failed!

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+    def serve_cars(self, car_collection:
+                   list) -> float:
+        filtered_cars = [car for car in car_collection
+                         if car.clean_mark < self.clean_power]
+        benefit = round(sum([car.comfort_class
+                        * ((self.clean_power - car.clean_mark)
+                            * self.average_rating)
+                        / self.distance_from_city_center
+                        for car in filtered_cars]), 1)
+        [self.wash_single_car(car) for car in filtered_cars]
```
vsmutok: DRY 

Here you use a loop three times; try to do it just once.

vsmutok (changes requested):

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_collection:
+                   list) -> float:
+        benefit = []
+        for car in car_collection:
+            if car.clean_mark < self.clean_power:
+                benefit.append(self.calculate_washing_price(car))
+                self.wash_single_car(car)
+        return round(sum(benefit), 1)
```
```suggestion
        benefit = 0
        for car in car_collection:
            if car.clean_mark < self.clean_power:
                benefit += self.calculate_washing_price(car)
                self.wash_single_car(car)
        return round(benefit, 1)
```
no need to use loop in this case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1834, opened 2025-01-29
Solution

chiper9:


app/main.py
```diff
@@ -1,8 +1,52 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            comfort_class: int,
+            clean_mark: int,
```
MNDonut, line 5: there are limits for these values, you should validate it. check the requirements

app/main.py
```diff
@@ -1,8 +1,52 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: int,
+            average_rating: float,
```
MNDonut, line 28: there are limits for these values, you should validate it. check the requirements

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        income = 0.0  # Доход от мойки автомобилей
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                # Рассчитываем стоимость мойки для этого автомобиля
+                cost = self.calculate_washing_price(car)
+                income += cost
```
MNDonut: or just:
income += self.calculate...
as you don't use 'cost' variable anywhere

MNDonut (changes requested):

Nattalli (approved):

PR #1846, opened 2025-02-02
Implemented Car and CarWashStation classes with car wash logic

Sotnikov-100:


app/main.py
```diff
@@ -1,8 +1,59 @@
+ffrom typing import List
```
kram3ko: mistake in word like "ff"

kram3ko (changes requested):
fix import

kram3ko (approved):

PR #1855, opened 2025-02-03
add Class Car with init method and 3 arguments,

ElfBerserk:
add class CarWashStation with method init and 4 arguments, add to class CarWashStation this methods:
serve_cars,calculate_washing_price,wash_single_car,rate_service edited by flake8

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,42 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,46 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        if distance_from_city_center == 0:
+            return
```
SerhiyTo: I recommend to raise the error, don't use return

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        income = 0
```
SerhiyTo: You can improve your code by using the `sum` function, without an extra iteration variable

SerhiyTo (changes requested):

PR #1861, opened 2025-02-06
Solution

BuryiOstap:


vsmutok (changes requested):
Tests Failed!

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+                 clean_power: int,
+                 average_rating: int,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list[Car]) -> int:
+        income = 0
```
SerhiyTo, line 23: If I were you, I would improve this function by replacing the loop with an additional variable to iterate with the use of the usual `sum()` function. This solution will cost us only 1 line of code and its readability will not deteriorate

SerhiyTo (approved):

app/main.py
BuryiOstap: def serve_cars(self, cars: list[Car]) -> int:
        return sum([self.wash_single_car(car) for car in cars])

 Something like this?

PR #1892, opened 2025-02-17
Develop Car wash station

AndyStarGitHub:
Car wash station


app/main.py
```diff
@@ -1,8 +1,54 @@
...
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> None:
+        cost = car.comfort_class
```
Yurkiv313, line 23: In my opinion, there's no need to store values in variables when you can access them directly. It's not a mistake, but I recommend showing the code to a mentor and getting their advice.

Yurkiv313 (approved):

PR #1894, opened 2025-02-17
Solution

iSevenpwnz:


app/main.py
```diff
@@ -1,8 +1,52 @@
...
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
+                total_income += price
+                self.wash_single_car(car)
+        return total_income
+
+    def rate_service(self, rating: float) -> None:
+        old_count = self.count_of_ratings
+        old_average = self.average_rating
```
Yurkiv313, line 48: You don't need to create the old_count and old_average variables, you can use their values in the formula.

Yurkiv313 (approved):

PR #1896, opened 2025-02-17
Implement Car and CarWashStation classes

PrimeGlorious:


app/main.py
```diff
@@ -1,8 +1,46 @@
...
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, car_list: list) -> float:
+        income = sum([self.calculate_washing_price(car)
+                      for car in car_list
+                      if car.clean_mark < self.clean_power])
```
Oneyura: You check the same conditions two times (line 26), you can use it one time.
Also, you use cycle "for" two time (line 25), but you can use once.

app/main.py
```diff
@@ -1,8 +1,46 @@
...
+
+        for car in car_list:
+            if car.clean_mark < self.clean_power:
+                self.wash_single_car(car)
+                self.rate_service(5.0)
```
Oneyura: In task, you must not set new service rate in func serve_cars

Oneyura (changes requested):
In task, you must not set new rating in func serve_cars
Also, you may optimize your code.

app/main.py
```diff
@@ -1,8 +1,44 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
```
oliasymak24, line 4: If you're breaking lines then each parameter must be from new one
Check everywhere

oliasymak24 (approved):

PR #2045, opened 2025-04-22
Solution

Nib20:


app/main.py
```diff
@@ -1,8 +1,43 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
MNDonut: it general it's better to always format parameters in this way:

    def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str
    ) -> None:

MNDonut (changes requested):

app/main.py
Nib20: thanks, I have fixed

app/main.py
```diff
@@ -32,11 +47,17 @@ def calculate_washing_price(self, car: Car) -> float:
...
+        car: Car
+    ) -> None:
         if self.clean_power > car.clean_mark:
             car.clean_mark = self.clean_power
 
-    def rate_service(self, new_rate: float) -> None:
+    def rate_service(
+        self,
+        new_rate: float
+    ) -> None:
```
MNDonut: one remark. do this formatting when the length of the line is bigger than 79 characters ;)
this is what PEP8 recommends

MNDonut (changes requested):

app/main.py
Nib20: fixed

MNDonut (approved):

PR #2153, opened 2025-06-09
Solution

PavloReutskiy:


app/main.py
```diff
@@ -1,8 +1,63 @@
...
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: Take into account that some attributes have an allowed range of values
image

app/main.py
```diff
@@ -1,8 +1,63 @@
...
+        total_income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
+                self.wash_single_car(car)
+                total_income += price
+        return round(total_income, 1)
+
+    def calculate_washing_price(self, car: Car) -> Union[float, int]:
```
MNDonut: use | operator in annotation instead

app/main.py
```diff
@@ -1,8 +1,63 @@
...
+        return round(washing_price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> None:
+        total_rating_sum = (self.count_of_ratings
+                            * self.average_rating
+                            + new_rating)
```
MNDonut: we usually recommend to move values on the next lines, so that parenthess are the only characters on their lines

MNDonut (changes requested):

MNDonut (approved):

PR #2156, opened 2025-06-11
Solution

ikocherovets:


app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut: Note that some attributes have a range of values

MNDonut (changes requested):

app/main.py
MNDonut: image

Losiev (changes requested):
accidentally appeared at DO board

PR #2177, opened 2025-06-19
Solution

Plovuezz:


app/main.py
```diff
@@ -1,8 +1,55 @@
...
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        overall_cost = ((((car.comfort_class
+                          * (self.clean_power - car.clean_mark))
+                         * self.average_rating))
+                        / self.distance_from_city_center)
```
Gleg0: In my opinion you should align the lines with the calculations.

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        overall_cost = ((((car.comfort_class
```
Gleg0: Why just replace overall_cost with return and don`t create it.

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+                         * self.average_rating))
+                        / self.distance_from_city_center)
+
+        overall_cost = round(overall_cost, 1)
+
+        return overall_cost
+
+    def serve_cars(self, cars: list) -> None:
+
+        income = 0
```
Gleg0, line 31: In my opinion you should use method sum with list comprehension in this function.It will be just oneliner.

app/main.py
```diff
@@ -1,8 +1,55 @@
...
+
+        for car in cars:
+            if self.clean_power > car.clean_mark:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def wash_single_car(self, car: Car) -> None:
+
+        car.clean_mark = self.clean_power
```
Gleg0, line 41: I think it`s not correct realization of this function becouse in README.txt has line "if wash_station.clean_power is greater than car.clean_mark".

Gleg0 (approved):

PR #2178, opened 2025-06-19
Solution

Gleg0:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+            ) / self.distance_from_city_center,
+            1
+        )
+
+    def wash_single_car(self, car: Car) -> float:
+        if self.clean_power > car.clean_mark:
+            income = self.calculate_washing_price(car)
+            car.clean_mark = self.clean_power
+            return income
+        return 0
```
Plovuezz, line 43: in my solution pytest consider this as an error, if it works here its great, because it more logical here

Plovuezz (approved):
looks good

PR #2185, opened 2025-06-21
Solution

Arch0998:


Vskesha (approved):
You solved this problem efficiently. Quality of code is good. Code is easy to understand. That's why approve.

PR #2189, opened 2025-06-22
Completed task

Salmon-eat:


app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
```
Artcrafterrra, line 10: According to the README, distance_from_city_center and clean_power should be of type float instead of int.

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, new_rating: float) -> float:
+        result_rating = (self.average_rating * self.count_of_ratings)
+        self.count_of_ratings += 1
+        self.average_rating = round((result_rating + new_rating)
+                                    / self.count_of_ratings, 1)
+        return self.average_rating
```
Artcrafterrra, line 38: return self.average_rating is unnecessary here. The method is supposed to modify the state, not return a result.

Artcrafterrra (approved):
Everything looks fine, but please pay attention to the comments and make sure there are no errors reported by flake8.

PR #2191, opened 2025-06-22
Implemented a solution

Otst0y:


DrGregoryHouse4676 (approved):
nice clean code, but I would make it even more readable

app/main.py
```diff
@@ -1,8 +1,67 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
+        if comfort_class not in range(1, 8):
+            self.comfort_class = 1
```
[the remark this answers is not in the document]
L1nk3rrr: Agreed with AI.

app/main.py
```diff
@@ -1,8 +1,67 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
L1nk3rrr: Better to pass every argument from a new line if you can't do it at one
```
    def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) -> None:
    ```

File: app/main.py [hunk 3]
```diff
@@ -1,8 +1,67 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float,
+                 clean_power: int, average_rating: float,
+                 count_of_ratings: int) -> None:
```
L1nk3rrr: same about formating

app/main.py
```diff
@@ -1,8 +1,67 @@
...
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        if 1 <= rate <= 5:
+            new_count_of_ratings = self.count_of_ratings + 1
+            new_average_rating = (
+                self.average_rating * self.count_of_ratings
+                + rate) / new_count_of_ratings
```
L1nk3rrr: Bad formatting, look for the checklist that was attached to the task at the end of the readme file

app/main.py
```diff
@@ -1,8 +1,67 @@
...
+        clean_diff = self.clean_power - car.clean_mark
+        if clean_diff <= 0:
+            return 0.0
+        washing_price = (
+            car.comfort_class
+            * clean_diff * self.average_rating
+            / self.distance_from_city_center
+        )
+
+        return round(washing_price, 1)
```
L1nk3rrr: Bad formatting, also washing_price looks like not a needed variable, as you can return your calculations at round.
Also, it would be better to practice a formatting your lines, as now it does not good from a readability side

L1nk3rrr (changes requested):

AnyoneClown (approved):
LGTM

PR #2192, opened 2025-06-22
Develop

mishahoshii:


app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark >= self.clean_power:
+            return 0.0
+
+        price = ((car.comfort_class
+                 *(self.clean_power
+                   - car.clean_mark)
+                 *self.average_rating)
```
SensibleN00B: add a space between * and self

app/main.py
```diff
@@ -1,8 +1,50 @@
...
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        if car.clean_mark >= self.clean_power:
+            return 0.0
+
+        price = ((car.comfort_class
+                 *(self.clean_power
```
SensibleN00B: add a space between * and (

SensibleN00B (changes requested):
File → Settings → Keymap → знайди "Reformat Code" → Ctrl+Alt+L
this is automatic code formatting so you don't have to manually enter spaces
and split the arguments in the init method so there is no error about a long string

app/main.py
mishahoshii: ok, thank you

app/main.py
mishahoshii: thank you

PR #2194, opened 2025-06-22
Solution

Sashoxxx:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
```
Salmon-eat: According to the README, clean_power should be of type float

Salmon-eat (approved):

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center
+        )
+        return round(price, 1)
```
L1nk3rrr: you don't need a variable, as you used it only 1 time.

Simply do:
```
return round(
    ....
    .....
    .....
    )

File: app/main.py [hunk 3]
```diff
@@ -1,8 +1,45 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int,
+                 clean_mark: int, brand: str) -> None:
```
L1nk3rrr: Better formating is when every param from new line if you can't write it into one line
```
    def __init__(
        self,
        comfort_class: int,
        clean_mark: int,
        brand: str,
    ) -> None:
    ```

File: app/main.py [hunk 4]
```diff
@@ -1,8 +1,45 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: float,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
L1nk3rrr: same about formating

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+    def serve_cars(self, cars: list) -> float:
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.wash_single_car(car)
+                car.clean_mark = self.clean_power
+        return round(income, 1)
+
+    def rate_service(self, mark: int) -> None:
+        if 0 <= mark <= 5:
```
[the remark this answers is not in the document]
L1nk3rrr: agree

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+            all_marks_sum = self.average_rating * self.count_of_ratings + mark
+            self.count_of_ratings += 1
+            self.average_rating =\
+                round(all_marks_sum / self.count_of_ratings, 1)
```
L1nk3rrr: Why is `\` used here? It looks like it's not needed, as it provides poor readability.

L1nk3rrr (changes requested):

app/main.py
Sashoxxx: > Why is `\` used here? It looks like it's not needed, as it provides poor readability.

app/main.py:48:80: E501 line too long (81 > 79 characters)
Moved "=" to next line for better readability

app/main.py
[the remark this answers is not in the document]
AnyoneClown: Fix this

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+                income += self.wash_single_car(car)
+                car.clean_mark = self.clean_power
+        return round(income, 1)
+
+    def rate_service(self, mark: int) -> None:
+        if 0 <= mark <= 5:
+            all_marks_sum = self.average_rating * self.count_of_ratings + mark
+            self.count_of_ratings += 1
+            self.average_rating \
+                = round(all_marks_sum / self.count_of_ratings, 1)
```
AnyoneClown:
```suggestion
            self.average_rating = round(
                all_marks_sum / self.count_of_ratings, 1
            )
```

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(
+            car.comfort_class
+            * (self.clean_power - car.clean_mark)
+            * self.average_rating
+            / self.distance_from_city_center, 1)
```
AnyoneClown:
```suggestion
        return round(
            car.comfort_class
            * (self.clean_power - car.clean_mark)
            * self.average_rating
            / self.distance_from_city_center, 1
        )
```

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #2196, opened 2025-06-23
Update main.py

B1g-Bear:


viannik (approved):
Overall, the implementation is correct and clean 👍
The only thing that could be improved is to add a check for boundary values in the __init__ method, since the task specification indicates the permissible ranges (for example, comfort_class must be from 1 to 7, clean_mark — from 1 to 10, etc.). This will help avoid potential errors in future use of the classes.

PR #2197, opened 2025-06-23
Solution

VadymKyrylko:


app/main.py
```diff
@@ -1,8 +1,47 @@
+#from typing import Callable
```
mishahoshii: you need to delete  '#'

app/main.py
```diff
@@ -1,8 +1,47 @@
...
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
```
mishahoshii, line 14: self.average_rating = round(average_rating, 1)

VadymKyrylko: Wait a minute, please.
Look at "new_average" above the self.average_rating = new_average.
I thought that I already did it. Do you agree?

mishahoshii: yes, you did it by your own way

mishahoshii (approved):

PR #2207, opened 2025-06-24
Solution

Bebaaaaa:


MeepoHustler:
👍

app/main.py
```diff
@@ -1,8 +1,52 @@
+from typing import List
+
+
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mrk: int, brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mrk
```
bohdankolisnyk04: clean_mrk -> clean_mark
I don't think saving one letter is worth writing it like that.

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+    def __init__(
+        self,
+        distance_from_city_center: float,
+        clean_power: int,
+        average_rating: float,
+        count_of_ratings: int,
+    ) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = float(average_rating)
```
bohdankolisnyk04: For `average_rating`, you're already annotating it as a float.

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: List[Car]) -> float:
+        total_income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
```
bohdankolisnyk04: 42. if car.clean_mark < self.clean_power:
but in method calculate_washing_price you also check  - if car.clean_mark >= self.clean_power:

bohdankolisnyk04: you no need var price, cause you can use:
total_income += self.calcutale_washing_price(car)

app/main.py
```diff
@@ -1,8 +1,52 @@
...
+                total_income += round(price, 1)
+                self.wash_single_car(car)
+        return round(total_income, 1)
```
bohdankolisnyk04: you alreade round you var total_income in 44 row

bohdankolisnyk04 (approved):
Overall, the code works and is easy to read, but there are a few points that can be simplified, which will improve the code. Some actions are performed twice in a row, even though this is absolutely unnecessary.

PR #2230, opened 2025-07-01
feat: solution

ArtemZakhar:
Hello)
Have a nice day)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -20,12 +20,7 @@ class CarWashStation:
         self.count_of_ratings = count_of_ratings
 
     def serve_cars(self, cars: list) -> float:
-        result = 0
-
-        for car in cars:
-            result += self.wash_single_car(car)
-
-        return result
+        return sum(self.wash_single_car(car) for car in cars)
 
     def wash_single_car(self, car: Car) -> float:
         if car.clean_mark > self.clean_power:
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,50 @@
 class Car:
-    # write your code here
+    def __init__(self,
+                 comfort_class: int,
+                 clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
     pass
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,49 @@
...
-    pass
+    def __init__(self,
+                 distance_from_city_center: float,
+                 clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
```
MNDonut, line 33: some attributes have allowed range of values
image

app/main.py
```diff
@@ -1,8 +1,49 @@
...
+
+    def serve_cars(self, cars: list) -> float:
+        result = 0
+
+        for car in cars:
+            result += self.wash_single_car(car)
+
+        return result
```
MNDonut: you can use `sum` function here

MNDonut (changes requested):

app/main.py
ArtemZakhar: @MNDonut so you want me to validate inputs or what?

app/main.py
ArtemZakhar: thank you!

app/main.py
MNDonut: It's the requirement of the task

MNDonut (changes requested):
You still should validate values that are specified in the requirements because, for example, you don't want users to have an ability to set rating not from 1 to 5, but from 1 to 999999999999 in your taxi application .

app/main.py
ArtemZakhar: @MNDonut sorry)

app/main.py
```diff
@@ -3,6 +3,11 @@ def __init__(self,
                  comfort_class: int,
                  clean_mark: int,
                  brand: str) -> None:
+        if comfort_class < 1 or comfort_class > 7:
```
oliasymak24: you can try to use this syntax
`if not 1 <= comfort_class <= 7:`

oliasymak24 (changes requested):
You correctly caught an idea but suggest use other syntax

app/main.py
ArtemZakhar: No way)
Thank you)

oliasymak24 (approved):

PR #2351, opened 2025-09-07
End 1

AlexFoxUA:


vsmutok (changes requested):
Why did you delete all the files (`.flake8`, `.github/workflows/test.yml`, `README.md`, and others)?
Please restore them, then complete the task and only after that submit it for review.

PR #2430, opened 2025-10-04
Develop

KuroMidori123:


app/main.py
```diff
@@ -1,8 +1,39 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
```
Dimosphen1, line 2: Place each argument including `self` on a new line in the future so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

Dimosphen1 (approved):
Looks good to me, well done! One comment is left for the future

PR #2470, opened 2025-10-23
Completed Task

ostapshpeha:
Succesfully tested by flake8 and pytest


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                price = self.calculate_washing_price(car)
+                self.wash_single_car(car)
+                income += price
+            else:
+                continue
```
stanislav-kishechko: This cycle is not necessary and will continue to work anyway.

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        return round(income, 1)
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
```
stanislav-kishechko, line 33: It is better to break the formula into several parts so that the code is clearer.

app/main.py
```diff
@@ -1,8 +1,45 @@
 class Car:
```
stanislav-kishechko, line 1: General recommendation to add docstrings for classes and functions

stanislav-kishechko (approved):

app/main.py
ostapshpeha: I appreciate your feedback! Thanks

PR #2479, opened 2025-10-25
added full classes Car and Carwash

AngryAndrii:
tests passed, flake8 clear, totally done

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
+        self.comfort_class = comfort_class
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        total_cost = 0
+        for car in cars:
+            print(car.clean_mark)
+            if car.clean_mark < self.clean_power:
+                total_cost += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return total_cost
+
+    def calculate_washing_price(self, car: Car) -> float:
+        sum_of_washing = ((car.comfort_class
+                          * (self.clean_power - car.clean_mark)
+                          * self.average_rating)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                          / self.distance_from_city_center)
+        return round(sum_of_washing, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
+
+    def rate_service(self, add_rating: float) -> None:
+        new_rating = (((self.average_rating
+                       * self.count_of_ratings) + add_rating)
+                      / (self.count_of_ratings + 1))
```
Iventyk: Missed one space before the "/" sign to align with the row above

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+                 count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = round(average_rating, 1)
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> float:
+        total_cost = 0
+        for car in cars:
+            print(car.clean_mark)
```
Iventyk: Do you really need "print" here?

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+        self.clean_mark = clean_mark
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: float, clean_power: int,
+                 average_rating: float,
+                 count_of_ratings: int) -> None:
```
Iventyk, line 16: If you’ve already moved an argument to a new line, I think all arguments should be placed on separate lines—one argument per line.

app/main.py
```diff
@@ -1,8 +1,42 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self, comfort_class: int, clean_mark: int,
+                 brand: str) -> None:
```
Iventyk, line 5: If you’ve already moved an argument to a new line, I think all arguments should be placed on separate lines—one argument per line.

app/main.py
```diff
@@ -1,8 +1,42 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        sum_of_washing = ((car.comfort_class
+                          * (self.clean_power - car.clean_mark)
+                          * self.average_rating)
+                          / self.distance_from_city_center)
+        return round(sum_of_washing, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        car.clean_mark = self.clean_power
```
Iventyk: You have to check if car.clean_mark < self.clean_power before changing attribute value according to requirements:

"wash_single_car - method, that washes a single car, so it should have clean_mark equals wash station's clean_power, if wash_station.clean_power is greater than car.clean_mark"

Iventyk (changes requested):

PR #2494, opened 2025-10-27
created classes Car and CarWashStation. For CWS was created 4 methods…

AlexMishKo:
… acourding to the task


dmytro-malyk-dm (approved):
Good solution!

PR #2528, opened 2025-11-06
Solution

Zikkolion:


app/main.py
```diff
@@ -1,8 +1,49 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
DaTrEvTeR, line 2: PEP8 tells us that we should use this code style.:
```py
class Car:
    def __init__(. # new line
        self,  # 1 tab
        comfort_class: int,
        clean_mark: int,
        brand: str,  # add a comma after the last argument and a new line
    ) -> None:
```

DaTrEvTeR (approved):
Well done!

PR #2618, opened 2025-11-29
Add files via upload

NiMiR13:
solve car wash station


alisahoys (approved):
Hi! Your solution is mostly correct, you just have to change the formula in the  calculate_washing_price method.
the corect formula: price = car.comfort_class * (self.clean_power - car.clean_mark) * self.average_rating / self.distance_from_city_center

PR #2705, opened 2026-01-12
Solution

IlyashenkoAleksandr:


IlyashenkoAleksandr:

IlyashenkoAleksandr:
Added a conditional check in wash_single_car

PR #2787, opened 2026-02-04
Solution

swirtam:


swirtam:
@mateacademy-ai-mentor firstly, the function serve_cars is supposed to return total income of all washed cars rounded to 1 decimal place. But Secondly, it sums the results of calls of calculate_washing_price functions and they are already rounded to 1 decimal. Since the sum of numbers, which are rounded to 1 decimal is by default rounded to 1 decimal, I think checking it again in serve_cars is not really necessary.

PR #2815, opened 2026-02-16
'Solution'

AlinaSv88:


app/main.py
```diff
@@ -1,8 +1,69 @@
...
+
+    def rate_service(self, rate: int) -> None:
+
+        self.average_rating = round(
+            (self.average_rating * self.count_of_ratings + rate)
+            / (self.count_of_ratings + 1), 1
+        )
+        self.count_of_ratings += 1
+
+        return
```
vsmutok: There is no need to return an empty `return`; you can simply remove it.

app/main.py
```diff
@@ -1,8 +1,69 @@
...
+        )
+
+        return price_per_car
+
+    def wash_single_car(self, car: Callable) -> None:
+
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+        return
```
vsmutok: There is no need to return an empty `return`; you can simply remove it.

vsmutok (changes requested):

app/main.py
AlinaSv88: done

DaTrEvTeR (approved):
Well done!

PR #2870, opened 2026-03-12
py-car-wash

andrearantes:
verify task


andrearantes (commented):
I made changes to the main.py code, please check, because on the Mate website, after I send a copy of the code, Luke responds that it is correct. I have already corrected the serve_cars method.

PR #2899, opened 2026-03-23
'Solution'

viakol-prog:


viakol-prog:
"Fix: correct calculate_washing_price guard and serve_cars logic"

viakol-prog (commented):
1

PR #2930, opened 2026-03-30
Solution

PaniShews:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+            return 0.0
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
+            car.clean_mark = self.clean_power
+
+    def serve_cars(self, cars: List[Car]) -> float:
```
andriihrechko, line 34: From Python 3.9+, use built-in lowercase types (list, dict, set) for annotations.

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        total_income = 0.0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
```
andriihrechko, line 37: Redundant car.clean_mark < self.clean_power checks should be removed from the wash_single_car and calculate_washing_price methods, as this condition is already validated in the serve_cars method.

andriihrechko (approved):

PR #3083, opened 2026-06-02
Add Car and CarWashStation classes with method

naderidkeidek-star:
3rd Try
Implemented Car and CarWashStation classes with methods for washing cars and calculating prices.
Pulled


naderidkeidek-star:
6666

PR #3145, opened 2026-06-28
auto washing cost calculation

DanZhul:


DanZhul:
@mateacademy-ai-mentor please recheck the PR

PR #3165, opened 2026-07-06
Car wash station implementation 

oleksii-tkachuk:


app/main.py
```diff
@@ -1,8 +1,54 @@
 class Car:
-    # write your code here
-    pass
+    def __init__(self,
```
DonDmytron, line 2: def __init__(
            self,
            comfort_class: int,
            clean_mark: int,
            brand: str
    ) -> None: ... - looks better

app/main.py
```diff
@@ -1,8 +1,54 @@
...
+             * (self.clean_power - car.clean_mark)
+             * self.average_rating)
+            / self.distance_from_city_center,
+            1
+        )
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power > car.clean_mark:
+            car.clean_mark = self.clean_power
+        return
```
DonDmytron, line 34: By default, functions and methods return `None` unless explicitly specified. The `return` statement is unnecessary here.

DonDmytron (approved):

PR #3181, opened 2026-07-14
Solution

darkkkCoDeR:


app/main.py
```diff
@@ -1,8 +1,84 @@
...
+
+        @:param cars: list of Car's
+        @:return income: float
+        """
+
+        income = 0.0
+        if len(cars) > 0:
+            for car in cars:
+                if self.clean_power > car.clean_mark:
+                    income += round(self.calculate_washing_price(car), 1)
```
VadimPusiak788: Rounding each car's price individually before summing causes compounding rounding error, then rounding the total again compounds it further. For example, several prices like 1.24, 1.24, 1.24 round to 1.2 each (sum 3.6), whereas the true sum 3.72 rounds to 3.7 — a 0.1 discrepancy that grows with more cars.

Suggest rounding only once, on the final total:

VadimPusiak788 (changes requested):

darkkkCoDeR:
fix compounding rounding error and code style:
    - Fixed Rounding Error: Moved round(..., 1) out of the loop to run only once on the final total, preventing accumulated precision loss.
    - Cleaned Logic: Removed redundant if len(cars) > 0 check.
    - Style Fix: Resolved flake8 visual indentation warnings.

VadimPusiak788 (approved):

PR #3186, opened 2026-07-17
Solved task

Danpsyper:


app/main.py
```diff
@@ -1,8 +1,36 @@
...
-    # write your code here
-    pass
+    def __init__(self, distance_from_city_center: int, clean_power: int,
+                 average_rating: float, count_of_ratings: int) -> None:
+        self.distance_from_city_center = distance_from_city_center
+        self.clean_power = clean_power
+        self.average_rating = average_rating
+        self.count_of_ratings = count_of_ratings
+
+    def serve_cars(self, cars: list) -> int:
```
Sn1kls, line 17: return type should be `-> float` (you return `round(..., 1)`). Optionally `cars: list[Car]`

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        income = 0
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+        return income
+
+    def calculate_washing_price(self, car: tuple) -> int:
```
Sn1kls, line 25: `car` is a `Car`, not a `tuple`
 And return type is `float`, not `int`

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        return round(car.comfort_class * (self.clean_power - car.clean_mark)
+                     * self.average_rating / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: tuple) -> None:
+        car.clean_mark = self.clean_power
```
Sn1kls, line 30: `car:Car` and add the condition from the task – only wash if `self.clean_power > car.clean_mark`

Sn1kls (changes requested):

PR #3241, opened 2026-08-07
solution

gyxx1dd:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        self.brand = brand
 
 
 class CarWashStation:
-    # write your code here
-    pass
+    def __init__(
+            self,
+            distance_from_city_center: float,
+            clean_power: float,
```
FedirKiiko, line 15: better annotate with int type, like in car.clean_mark

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+        return round(
+            cars.comfort_class
+            * (self.clean_power - cars.clean_mark)
+            * self.average_rating / self.distance_from_city_center,
+            1,
+        )
+
+    def serve_cars(self, cars: Car) -> float:
+        result_price = 0
+        for i in cars:
```
FedirKiiko, line 34: better use "car" instead of "i"

FedirKiiko (approved):

PR #3245, opened 2026-08-09
Solution

AndrewBudkin1:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def serve_cars(self, cars: list[Car]) -> float:
+        income = 0
+
+        for car in cars:
+            if car.clean_mark < self.clean_power:
+                income += self.calculate_washing_price(car)
+                self.wash_single_car(car)
+
+        return round(income, 1)
```
CemXokenc, line 30: I think we can do without that

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def calculate_washing_price(self, car: Car) -> float:
+        price = (car.comfort_class * (self.clean_power - car.clean_mark)
+                 * self.average_rating / self.distance_from_city_center)
+
+        return round(price, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if car.clean_mark < self.clean_power:
```
CemXokenc, line 39: We are already calling this method after such a check

CemXokenc:
But overall, everything is fine

CemXokenc (approved):

PR #3268, opened 2026-08-10
Imlementation of car washing station

OkeNNV:


app/main.py
```diff
@@ -1,8 +1,43 @@
...
+        return income
+
+    def calculate_washing_price(self, car: Car) -> float:
+        return round(car.comfort_class
+                     * (self.clean_power - car.clean_mark)
+                     * self.average_rating
+                     / self.distance_from_city_center, 1)
+
+    def wash_single_car(self, car: Car) -> None:
+        if self.clean_power >= car.clean_mark:
```
NikitaFilipovWork, line 35: Just ">" is better. Because if it is 5 and 5, we will rewrite it which is unnecessary

app/main.py
```diff
@@ -1,8 +1,43 @@
...
+            car.clean_mark = self.clean_power
+
+    def rate_service(self, rate: int) -> None:
+        print(self.average_rating)
```
NikitaFilipovWork: Forgot to delete print

NikitaFilipovWork (approved):
All good)

